PR #15, opened 2019-02-19
chore(sec2kes): script to convert a Secret to an ExternalSecret + AWS SM entries

silasbw:
You would do something like:

```
kubectl -n simple-service get secret simple-service -o json > simple-service.json
mkdir simple-service-things
node bin/sec2kes.js --input simple-service.json --ouput simple-service-things
```

And then follow the instructions logged to the console. They would be something like:

```
aws secretsmanager create-secret  --name simple-service/service-account --secret-string file://simple-service-things/service-account.json
aws secretsmanager create-secret  --name simple-service/magic-api-cert --secret-string file://simple-service-things/magic-api-cert.json
kubectl apply -f simple-service-things/external-secret.yaml
```

### DIFF
--- a/bin/sec2kes.js
+++ b/bin/sec2kes.js
@@ -1,0 +1,110 @@
+/* eslint-disable no-sync, no-console */
+
+const fs = require('fs');
+const path = require('path');
+const yaml = require('js-yaml');
+
+function basenameGrouping(input) {
+  const data = input.data;
+  const grouped = Object.keys(data).reduce((awsSecrets, name) => {
+    const base = name.split('.').slice(0, -1).join('.');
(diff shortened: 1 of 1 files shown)


silasbw:
meh, I don't know if this belongs here or at `silasbw/` or somewhere else, but I thought it was worth sharing and getting feedback on.

JacopoDaeli:
@silasbw I like this. Let's think how to make split out. Maybe a `kes-cli` module?

jcrugzz:
@silasbw yea a way to make this easily automaetable seems legit. 👍

silasbw:
See https://github.com/silasbw/external-secrets-cli

PR #16, opened 2019-02-19
chore(manifest): add a tag external-secrets.yml

silasbw:


external-secrets.yml
```diff
@@ -69,6 +69,6 @@ spec:
     spec:
       serviceAccountName: kubernetes-external-secrets-service-account
       containers:
-        - image: "godaddy/kubernetes-external-secrets"
+        - image: "godaddy/kubernetes-external-secrets:5ed5188"
```
JacopoDaeli: Shouldn't this be `latest`?

silasbw: It's currently latest (eliding a tag defaults to latest).  I'm changing from latest to a specific build so it's explicit which version of kubernetes-external-secrets is running.

PR #19, opened 2019-02-20
chore(image): bump tag in external-secrets.yml

silasbw:


silasbw:
Rote change / should be part of CICD.  Merging.

PR #27, opened 2019-03-08
fix(backends): fix secretsManager backend name

JacopoDaeli:
NEED TO CREATE NEW DOCKER IMAGE AND SEPARATE PR TO UPDATE TAG IN DEPLOYMENT FILES.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ kind: ExternalSecret
 metadata:
   name: hello-service
 secretDescriptor:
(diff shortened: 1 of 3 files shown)


JacopoDaeli:
@silasbw yes, let's do that: https://github.com/godaddy/kubernetes-external-secrets/pull/27/files#diff-c8c5ee8e781f831e4927a392fa3565e6R42

Issue #52, opened 2019-05-01
Create a Helm Chart


davidholsgrove:
To make installation of the kubernetes-external-secrets controller and related resources a bit easier, a helm chart can be added to this repository in the first instance (and hopefully promoted to https://github.com/helm/charts/tree/master/incubator for public consumption).

This Helm Chart can also allow easy configuration of the environment variables for the container, like AWS_REGION and the events / poller intervals (addresses https://github.com/godaddy/kubernetes-external-secrets/issues/51).

PodAnnotations can also be used to configure kube2iam role's to grant access to AWS Secret Manager for example.

Issue #51, opened 2019-04-30
AWS Region should be required


kenske:
The config defaults to `us-west-2` for the AWS region which seems arbitrary and caused some pain while troubleshooting. Instead, the AWS region should be set explicitly in the deployment. The documentation should be updated to specify where to modify the value if needed.

JacopoDaeli:
Created a PR to fix this issue. Adding a `helm` chart is going to be the next step for simplifying how people can run the daemon with custom config.

PR #53, opened 2019-05-01
Add helm chart for kubernetes-external-secrets

davidholsgrove:
https://github.com/godaddy/kubernetes-external-secrets/issues/52

Set default values for AWS_REGION and poller / event intervals, which can be overridden by `--set` during helm installation, e.g;

```
helm install --name kube-external-secrets \
--namespace kube-secrets \
--set deployment.aws.region="ap-southeast-2" \
--set deployment.interval.poller="600000" \
--set podAnnotations."iam\.amazonaws\.com/role"=k8s-secretmanager-role \
charts/kubernetes-external-secrets
```

If PR accepted, this repository could use github pages to 'host' the helm chart initially for people to install directly (example: https://github.com/argoproj/argo-helm)


charts/kubernetes-external-secrets/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+name: kubernetes-external-secrets
+version: 0.1.0
```
JacopoDaeli: Should the `helm` chart version match the `appVersion`?

charts/kubernetes-external-secrets/Chart.yaml
```diff
@@ -0,0 +1,9 @@
...
+appVersion: 1.2.0
+description: Kubernetes External Secrets CustomResourceDefinition
+keywords:
+  - kubernetes-external-secrets
+  - secrets
+home: https://github.com/godaddy/kubernetes-external-secrets
```
JacopoDaeli: missing newline at the end of the file

charts/kubernetes-external-secrets/templates/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+    spec:
+      serviceAccountName: {{ template "kubernetes-external-secrets.fullname" . }}
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+          env:
+            - name: AWS_REGION
```
JacopoDaeli: After https://github.com/godaddy/kubernetes-external-secrets/pull/51, we should remove `AWS_REGION` in favor of `AWS_DEFAULT_REGION`.
We should also add `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, for people that runs the daemon in a non-AWS environment and need to explicitly set AWS credentials (since IAM roles for ec2 and eks are not available).

charts/kubernetes-external-secrets/templates/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+              value: "{{ .Values.deployment.aws.region }}"
+            - name: EVENTS_INTERVAL_MILLISECONDS
+              value: "{{ .Values.deployment.interval.events }}"
+            - name: POLLER_INTERVAL_MILLISECONDS
+              value: "{{ .Values.deployment.interval.poller }}"
+{{- if .Values.extraEnv }}
```
JacopoDaeli: Let's just use `.Values.env`.

charts/kubernetes-external-secrets/templates/deployment.yaml
JacopoDaeli: Also, let's be more flexible here and simply use `{{ toYaml .Values.env | indent 12 }}` instead.

charts/kubernetes-external-secrets/templates/rbac.yaml
```diff
@@ -0,0 +1,43 @@
...
+    heritage: {{ .Release.Service }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "kubernetes-external-secrets.fullname" . }}
+subjects:
+  - name: {{ template "kubernetes-external-secrets.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace | quote }}
+    kind: ServiceAccount
+{{- end -}}
```
JacopoDaeli: missing end of file

charts/kubernetes-external-secrets/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: ServiceAccount
+metadata:
+  name: {{ template "kubernetes-external-secrets.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    app: {{ template "kubernetes-external-secrets.name" . }}
+    chart: {{ template "kubernetes-external-secrets.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+{{- end -}}
```
JacopoDaeli: missing end of file

charts/kubernetes-external-secrets/values.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+image:
+  repository: godaddy/kubernetes-external-secrets
+  tag: 1.2.0
+  pullPolicy: IfNotPresent
+
+nameOverride: ""
+fullnameOverride: ""
+
+extraEnv: {}
```
JacopoDaeli: let's just use `env : {}`

charts/kubernetes-external-secrets/values.yaml
```diff
@@ -0,0 +1,53 @@
+# Default values for kubernetes-external-secrets.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+deployment:
```
JacopoDaeli: Let's just use `env` instead. remove `deployment.

charts/kubernetes-external-secrets/templates/deployment.yaml
JacopoDaeli: Also, prefer `nindent` over `indent`, which is available since helm 2.7: https://github.com/helm/helm/issues/3854 - it makes templates easier to read and maintain.

charts/kubernetes-external-secrets/Chart.yaml
davidholsgrove: Having a separate chart `version` from the `appVersion` lets us iterate and bump the chart content without changing the version of the app itself: https://github.com/helm/helm/blob/master/docs/charts.md#charts-and-versioning

charts/kubernetes-external-secrets/Chart.yaml
davidholsgrove: Damn, `helm create` generated all the source templates, will add some newlines 👍

charts/kubernetes-external-secrets/templates/deployment.yaml
davidholsgrove: Yeah, fair point. This PR was a minimal, helm chart only addition without changes to the application. I just exposed a mechanism to set the env variables already been looked for in `config/environment.js`.

`nindent` is cool, didnt know about that addition 👍

charts/kubernetes-external-secrets/Chart.yaml
JacopoDaeli: Thanks for pointing this out. Totally agree with it! Let' set `version: 1.0.0` then.

charts/kubernetes-external-secrets/Chart.yaml
JacopoDaeli: Eh eh, I run into this once too :D Github is pretty good to catch this :)

charts/kubernetes-external-secrets/templates/deployment.yaml
davidholsgrove: While its not in the Helm best practice guide, using `extraEnv` and `extraArgs` is a (I believe) a fairly standard approach to expose a consistent place to allow additional values not explicitly required or covered by the chart author (https://github.com/helm/charts/search?q=extraEnv&unscoped_q=extraEnv)

Then by setting the env variables that *must* be provided in the `deployment.yaml` container spec and `values.yaml`, we're guiding the user to what needs to be looked at?

JacopoDaeli: Yes I understand your point. However, I do not see, especially at this moment, any use case for it: when do we actually need to set an "extra env" beside standard env stuff such as `AWS_*` and `INTERVAL_*` ?
Finally, if there is any, I think it's still ok to simply add it to `.Values.env` to have 1:1 mapping. It's just more clear. What do you think?

davidholsgrove: Yeah fair point - how about we follow the helm chartmuseum's own approach?

https://github.com/helm/charts/blob/master/stable/chartmuseum/templates/deployment.yaml#L39-L43
https://github.com/helm/charts/blob/master/stable/chartmuseum/values.yaml#L10-L15

That way the `values.yaml` will be a simpler;
```
env:
  AWS_REGION: "us-west-2"
  EVENTS_INTERVAL_MILLISECONDS: 60000
  POLLER_INTERVAL_MILLISECONDS: 10000
```

JacopoDaeli: I am not sure I understand what's different here versus what I previously suggested.

JacopoDaeli: You can also commit and I'll review again.

davidholsgrove: 👍 Theres two approaches to setting env vars in `values.yaml` I guess - the alternative would be something like;
```
env:
  - name: "AWS_REGION"
    value: "us-west-2"
```
I'll push through a new commit with the Chart Museum method

charts/kubernetes-external-secrets/templates/deployment.yaml
davidholsgrove: Going to mark this as resolved - changes to the application to look for additional AWS_* env vars can be addressed in a separate PR.

silasbw:
Can you update the README.md with the helm-based install instructions?

davidholsgrove:
@silasbw - no worries, will add a section on the README.md with expectation the user will initially clone the repo, and run the helm install against the local copy of the chart. That instruction can be updated at a later stage if the chart is packaged and 'released' as a repo 👍

davidholsgrove:
Hi @JacopoDaeli & @silasbw - let me know if you want any further changes made before this can be merged?

thanks,
David

charts/kubernetes-external-secrets/templates/deployment.yaml
```diff
@@ -0,0 +1,51 @@
...
+    spec:
+      serviceAccountName: {{ template "kubernetes-external-secrets.fullname" . }}
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+          env:
+          {{- range $name, $value := .Values.env }}
```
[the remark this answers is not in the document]
jcrugzz: @JacopoDaeli that would put the burden on the consumer to make an object with name and value parameters rather than having an `env.Key=value` situation which this creates. I think this is better for the consumer.

README.md
```diff
@@ -34,6 +34,59 @@ kubectl apply -f https://raw.githubusercontent.com/godaddy/kubernetes-external-s
...
+helm delete kubernetes-external-secrets
+```
+
+#### Configuration
+
+The following table lists the configurable parameters of the `kubernetes-external-secrets` chart and their default values.
+
+| Parameter                            | Description                                                  | Default                                                 |
+| ------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------- |
+| `env.AWS_DEFAULT_REGION`             | Set AWS_DEFAULT_REGION in Deployment Pod                     | `us-west-2`                                             |
```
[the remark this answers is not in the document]
jcrugzz: The `kubernetes-external-secrets` service itself supports certain env vars including all the aws specific ones. I think here we should just indicate that we support passing in any ENV var and not be so specific in the helm docs. It should just be clear that helm the ENV vars for the service would be specified via the `env` object in a particular way as you have defined it.

charts/kubernetes-external-secrets/templates/deployment.yaml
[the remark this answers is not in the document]
JacopoDaeli: Ok it makes sense. I am ok with leaving it in this way.

neuromantik33:
Sorry to be nit picky, but this helm chart is quite unsable as is. I'm not running in AWS nor do I use vault but I'm unable to remove the environment variables referencing the latter since they are merged with the default `values.yaml`. Is there any way the helm chart can be more _agnostic_ as to the secret providers? Or perhaps provide a simple raw manifests with kustomization that way we can bypass this chart alltogether? I'm open for suggestions

Flydiverny:
> Sorry to be nit picky, but this helm chart is quite unsable as is. I'm not running in AWS nor do I use vault but I'm unable to remove the environment variables referencing the latter since they are merged with the default `values.yaml`. Is there any way the helm chart can be more _agnostic_ as to the secret providers? Or perhaps provide a simple raw manifests with kustomization that way we can bypass this chart alltogether? I'm open for suggestions

Please open a new issue or make a PR for improvements :)
I don't really see how its an issue with the env vars tho, whether an env var is set shouldn't make or break your day? Whats the real issue you are facing? To get raw manifests you can use [helm template](https://helm.sh/docs/helm/helm_template/).

PR #60, opened 2019-05-09
Fix and extend documentation

napcae:
A few improvements to the documentation, things that were unclear to me as and outsider of this project:
* You need to set AWS_REGION instead of AWS_DEFAULT_REGION, this is specified in the official docs as well: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html 
* the controller needs either credentials or a role to read from AWS Secrets Manager, it's kinda obvious but I feel like this belongs in the doc anyway
* small nitpick since it's not mentioned anywhere but supported, adding namespace to the example as an implicit option

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,28 @@ data:
   password: MTIzNA==
 ```
 
+The controller needs to access AWS Secrets Manager via a role. How depends on your cluster setup, [kube2iam](https://github.com/jtblin/kube2iam) and [kiam](https://github.com/uswitch/kiam) are the most common
+
+```
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "secretsmanager:GetRandomPassword",
+                "secretsmanager:GetResourcePolicy",
+                "secretsmanager:GetSecretValue",
+                "secretsmanager:DescribeSecret",
+                "secretsmanager:ListSecrets",
+                "secretsmanager:ListSecretVersionIds"
+            ],
+            "Resource": "*"
+        }
+    ]
+}
+```
+
 ## Backends
 
 kubernetes-external-secrets supports only AWS Secrets Manager.
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ apiVersion: 'kubernetes-client.io/v1'
 kind: ExternalSecret
 metadata:
   name: hello-service
+  namespace: default
 secretDescriptor:
   backendType: secretsManager
   data:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,7 +61,7 @@ The following table lists the configurable parameters of the `kubernetes-externa
 
 | Parameter                            | Description                                                  | Default                                                 |
 | ------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------- |
-| `env.AWS_DEFAULT_REGION`             | Set AWS_DEFAULT_REGION in Deployment Pod                     | `us-west-2`                                             |
```
JacopoDaeli: we got rid of this already with the last merged PR https://github.com/godaddy/kubernetes-external-secrets/commit/05b92a2aa5f648f9269bbf8f8a16cffa6bb5f998#diff-04c6e90faac2675aa89e2176d2eec7d8L64. Please sync your fork with our latest master.

README.md
```diff
@@ -102,6 +102,7 @@ apiVersion: 'kubernetes-client.io/v1'
 kind: ExternalSecret
 metadata:
   name: hello-service
+  namespace: default
```
JacopoDaeli: No need for this.

README.md
```diff
@@ -133,6 +134,28 @@ data:
   password: MTIzNA==
 ```
 
+The controller needs to access AWS Secrets Manager via a role. How depends on your cluster setup, [kube2iam](https://github.com/jtblin/kube2iam) and [kiam](https://github.com/uswitch/kiam) are the most common IAM role plugins.
```
JacopoDaeli: Maybe I am wrong, but plugins are not necessary when your cluster runs within AWS. Eg, https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html

Can you clarify this?

README.md
JacopoDaeli: image

README.md
silasbw: Some EKS users use kube2iam or another project to implement IAM roles for `Pods`.  We expect native EKS support soon, so we haven't bothered to run kube2iam or one of the other projects.

README.md
napcae: Hey, I added this so it is more clear that you can also specify another namespace: 
>small nitpick since it's not mentioned anywhere but supported, adding namespace to the example as an implicit option

README.md
napcae: Since you've published this as open source, I wouldn't assume that everyone is running EKS(wasn't clear to me either from this project description/doc/readme); some are running their own clusters on top of EC2 in which case you'd need this. I could rephrase it. Wdyt?

README.md
napcae: Will do, what happens if you don't declare a variable for `AWS_REGION`? Because you have removed this entirely in master.

README.md
JacopoDaeli: I'd keep this out and still give the user the possibility to install this in a whatever namespace using the `--namespace` arg.

README.md
JacopoDaeli: This is AWS specific and I'd keep this out of the project documentation.

JacopoDaeli:
@napcae thanks for this PR. Closing without merging. Please see comments above.

Issue #73, opened 2019-05-30
Feature request: ability to generate resulting secrets in a file format


darose:
In the example provided in your readme, the resulting secret generated by an external secret looks like so:

```
apiVersion: v1
kind: Secret
metadata:
  name: hello-service
type: Opaque
data:
  password: MTIzNA==
```

However, if you try to mount this secret object into a pod, what will happen is that the secret will get mounted as a directory, containing a file called password.  If there were 2 keys that you pulled from ASM - e.g., username and password - then the directory would contain 2 files, username, and password.

This isn't really ideal from a usability standpoint.  It would be preferable to generate a single file, say config.txt, that contains the 2 keys and values inside of it.  E.g.:
```
username=foo
password=bar
```

This can accomplished with Kubernetes secrets, by using the "stringData" functionality as described [here](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually) E.g.:
```
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
type: Opaque
stringData:
  config.yaml: |-
    apiUrl: "https://my.api.com/api/v1"
    username: {{username}}
    password: {{password}}
```

This would make for a better use case, as it would then be possible to just mount a single file (config.yaml in this case) into a container.

For bonus points, it would additionally be nice to be able to specify the format of the resulting file - e.g., txt, json, etc.

JacopoDaeli:
@darose can you describe better why "This isn't really ideal from a usability standpoint" ?

FYI, you can easily accomplish this already. Simply create your "config.yaml" secret in secret manager directly.

JacopoDaeli:
Result would be:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
type: Opaque
data:
  config.yaml: <base64encoded>
```

JacopoDaeli:
Does it make sense @darose? If yes, I will close the issue.

darose:
When I say it isn't ideal from a usability standpoint, I mean that in the sense that it's preferable to provide an application with a single file of secrets containing multiple key value pairs, rather than multiple files in a directory, each containing a single value.

[...]

PR #74, opened 2019-06-01
fix(AWSSM): treat value as object iff the es specifies .property

silasbw:
See https://github.com/godaddy/kubernetes-external-secrets/issues/73 for
discussion.

### DIFF
--- a/lib/backends/kv-backend.js
+++ b/lib/backends/kv-backend.js
@@ -28,8 +28,16 @@ class KVBackend extends AbstractBackend {
       const value = await this._get({ secretKey: secretProperty.key })
 
       if ('property' in secretProperty) {
-        return value[secretProperty.property]
+        let parsedValue
+        try {
+          parsedValue = JSON.parse(value)
+        } catch (err) {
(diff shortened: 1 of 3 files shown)


lib/backends/kv-backend.js
```diff
@@ -28,8 +28,16 @@ class KVBackend extends AbstractBackend {
       const value = await this._get({ secretKey: secretProperty.key })
 
       if ('property' in secretProperty) {
-        return value[secretProperty.property]
+        let parsedValue
+        try {
+          parsedValue = JSON.parse(value)
+        } catch (err) {
+          this._logger.warn(`Failed to JSON.parse '${value}':`, err)
+          return undefined
```
JacopoDaeli: no need to return `undefined`, just `return this._logger.warn...)` instead ?

silasbw: If we do that then this code takes a dependency on the return value of whatever logger we're using.

I'd rather have an additional line of code than an additional dependency on the return values of whatever logger we're using.

JacopoDaeli: Makes sense, can we just `return` then? `undefined` should be implicit.

Issue #46, opened 2019-04-16
Avoid putting secrets to ETCD


max-lobur:
Current implementation creates an etcd object with base64 encoded secret, which may potentially leak later. 

Allow to put secrets to a volume, e.g. add a mutation webhook which adds an initContainer, which fetches & writes secrets to a volume shared with a container.

Pros: 
 - no secrets in etcd
 - iam role can be granularly assigned to each initContainer (based on namespace annotation for example) when using https://github.com/uswitch/kiam , no need to give access to all secrets to the controller

silasbw:
Hey @max-lobur, how much does using [encryption at rest](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/) mitigate your concerns around etcd storage?

max-lobur:
- It's just about hacked/dumped eventually. 
- Depending of https://coreos.com/etcd/docs/latest/op-guide/configuration.html#--auto-compaction-retention etcd even can have values of deleted secrets. 
- Etcd (and k8s) backups must be treated carefully because now they contain sensitive data.

silasbw:
Thanks, that's helpful for understanding the potential priority of an addition like this.

silasbw:
I did some planning for this and some initial implementation work:

- [x] [POC Init Container example](https://github.com/godaddy/kubernetes-external-secrets/pull/77)
- [ ] [Issue to discuss / plan what the "API" is](https://github.com/godaddy/kubernetes-external-secrets/issues/78)
- [ ] [Implementation work to get the admission webhook server up](https://github.com/godaddy/kubernetes-external-secrets/issues/79)
- [ ] [Tie it all together end-to-end](https://github.com/godaddy/kubernetes-external-secrets/issues/81)

Feel free to jump in a give feedback (especially on https://github.com/godaddy/kubernetes-external-secrets/issues/78) or help out with coding :stuck_out_tongue:

gazal-k:
> * It's just about etcd hacked/dumped eventually.
> * Depending of https://coreos.com/etcd/docs/latest/op-guide/configuration.html#--auto-compaction-retention etcd even can have values of deleted secrets.
> * Etcd (and k8s) backups must be treated carefully because now they contain secrets.

[...]

PR #77, opened 2019-06-02
feat(bin/poll): CLI for one-shot polling and writing to files

silasbw:
Inlcude an example of using `npm run poll` as an Init Container. Part of the
design proposed here:

https://github.com/godaddy/kubernetes-external-secrets/issues/46

### DIFF
--- a/bin/poll.js
+++ b/bin/poll.js
@@ -1,0 +1,50 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const path = require('path')
+
+const {
+  backends,
+  customResourceManifest,
+  kubeClient,
+  logger
+} = require('../config')
+
+async function main () {
+  const destinationPath = process.env['DESTINATION_PATH']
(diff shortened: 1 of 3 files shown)


bin/poll.js
```diff
@@ -0,0 +1,50 @@
...
+  backends,
+  customResourceManifest,
+  kubeClient,
+  logger
+} = require('../config')
+
+async function main () {
+  const destinationPath = process.env['DESTINATION_PATH']
+  const externalSecretName = process.env['EXTERNAL_SECRET_NAME']
+  const namespace = process.env['EXTERNAL_SECRET_NAMESPACE']
```
jeffpearce: Would it be worth throwing an exception if any of these aren't provided, to help debug (by offering a specific error message)?

JacopoDaeli:
@silasbw do you still want to merge this work? If yes, I can review.

Issue #63, opened 2019-05-16
Failure while polling the secrets


kenske:
I'm trying to get kubernetes-external-secrets to work, but it's currently failing and the error logs don't seem to show any useful clues as to why it's happening:

```
2019-05-16T00:22:18.787878009Z {"level":30,"time":1557966138787,"pid":16,"hostname":"kubernetes-external-secrets-5bd4b59b5d-bf8gj","msg":"running poll","v":1}
2019-05-16T00:22:18.78823308Z {"level":30,"time":1557966138788,"pid":16,"hostname":"kubernetes-external-secrets-5bd4b59b5d-bf8gj","msg":"fetching secret property github_access_token","v":1}
2019-05-16T00:22:18.882898059Z {"level":50,"time":1557966138879,"pid":16,"hostname":"kubernetes-external-secrets-5bd4b59b5d-bf8gj","msg":"failure while polling the secrets {}","v":1}
2019-05-16T00:22:28.791813549Z {"level":30,"time":1557966148791,"pid":16,"hostname":"kubernetes-external-secrets-5bd4b59b5d-bf8gj","msg":"running poll","v":1}
2019-05-16T00:22:28.791997103Z {"level":30,"time":1557966148791,"pid":16,"hostname":"kubernetes-external-secrets-5bd4b59b5d-bf8gj","msg":"fetching secret property github_access_token","v":1}
2019-05-16T00:22:28.797970739Z {"level":30,"time":1557966148797,"pid":16,"hostname":"kubernetes-external-secrets-5bd4b59b5d-bf8gj","msg":"stopping and removing poller github-access-token_9063402","v":1}
2019-05-16T00:22:28.798018978Z {"level":30,"time":1557966148797,"pid":16,"hostname":"kubernetes-external-secrets-5bd4b59b5d-bf8gj","msg":"stopping poller","v":1}
```

Any ideas?

I ssh'd into the pod, created a nodejs test script that fetches the secret, and it does work.

JacopoDaeli:
@kenske I am going to look into this. Can you share the `content/structure` of your secret (remember to mask the token)? The `{}` is really weird. That should be a serialized error.

michas2:
I ran into exactly the same problem right now.
I debugged it to this line: https://github.com/godaddy/kubernetes-external-secrets/blob/master/lib/backends/kv-backend.js#L57

There you find: `Buffer.from(secretPropertyValues[index], 'utf8')`
[...]

PR #87, opened 2019-06-06
Show error on missing property

kenske:
The polling function was not outputting an error message when a property is missing from a secret. This addresses #63 .

### DIFF
--- a/lib/backends/kv-backend.js
+++ b/lib/backends/kv-backend.js
@@ -35,6 +35,11 @@ class KVBackend extends AbstractBackend {
           this._logger.warn(`Failed to JSON.parse '${value}':`, err)
           return
         }
+
+        if (!(secretProperty.property in parsedValue)) {
+          throw new Error('Could not find property ' + secretProperty.property + ' in ' + secretProperty.key)
+        }
+
         return parsedValue[secretProperty.property]
       }
 
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -64,7 +64,8 @@ class Poller {
         return this._upsertKubernetesSecret({ secretDescriptor })
       }))
     } catch (err) {
-      this._logger.error('failure while polling the secrets', err)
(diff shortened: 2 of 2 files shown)


lib/backends/kv-backend.js
```diff
@@ -35,6 +35,11 @@ class KVBackend extends AbstractBackend {
           this._logger.warn(`Failed to JSON.parse '${value}':`, err)
           return
         }
+
+        if (!(secretProperty.property in parsedValue)) {
+          throw new Error('Could not find property ' + secretProperty.property + ' in ' + secretProperty.key)
```
jeffpearce: Thanks for the pul request! Could you add a unit test for this case?

Also, it's probably better to check `parsedValue[secretProperty.property]` to match the rest of the code, or even check the result on line 43 before returning.

kenske: Not sure I understand... `parsedValue[secretProperty.property]` could contain an empty value, so checking that wouldn't help with verifying if the property exists.

lib/poller.js
```diff
@@ -64,7 +64,8 @@ class Poller {
         return this._upsertKubernetesSecret({ secretDescriptor })
       }))
     } catch (err) {
-      this._logger.error('failure while polling the secrets', err)
+      this._logger.error('failure while polling the secrets')
+      this._logger.error(err.toString())
```
silasbw: nit: `this._logger.error(err)` should work

lib/backends/kv-backend.js
silasbw: nit
```js
throw new Error(`Could not find property ${secretProperty.property} in ${secretProperty.key}`)
```

silasbw:
@kenske we just pushed version `1.2.3` that includes your fix.  Thanks for tracking down the bug and sending in a fix.  :100: :1st_place_medal:

Issue #78, opened 2019-06-02
Volume front end API


silasbw:
kubernetes-external-secrets currently has a single "front end" that writes backend data to `Secret` objects (and `Pod`s manipulate the `Secret` objects in the usual ways).  The goal of this issue is to define the API for a new frontend that wrties data to a `Volume`. This is part of the effort to [avoid putting Secrets in ETCD](https://github.com/godaddy/kubernetes-external-secrets/issues/46).

The Volume frontend API must allow engineers to declare how kubernetes external secrets writes external secret data to a `Volume`. One approach is to use annotations on the `Pod` to identify which volumes in `.spec.volumes` represent an `ExternalSecret`:

```yaml
kind: Pod
metadata:
  annotations:
    externalsecrets.kubernetes-client.io/volume/db-secrets: 'true'
    externalsecrets.kubernetes-client.io/volume/client-secrets: 'true'
spec:
  containers:
  - name: test
    image: busybox
    volumeMounts:
      - name: db-secrets
        mountPath: /db-secrets
      - name: client-secrets
        mountPath: /client-secrets
      - name: other-stuff
        mountPath: /stuff
  volumes:
  - name: db-secrets
    emptyDir:
      medium: "Memory"
  - name: client-secrets
    emptyDir:
      medium: "Memory"
  - name: other-stuff
    configMap:
      name: stuff-config
```

We should discuss the pros and cons of an approach like this and discuss other potential approaches.

silasbw:
I think we have an OK first pass at this: https://github.com/godaddy/kubernetes-external-secrets/blob/master/api.md

PR #91, opened 2019-06-09
docs(frontends): propose a volume-based "frontend"

silasbw:
Propose and document an API for writing secret data to volumes.

See: https://github.com/godaddy/kubernetes-external-secrets/issues/78

### DIFF
--- a/api.md
+++ b/api.md
@@ -87,7 +87,7 @@ secretDescriptor:
 ```
 
 The value of `externalsecrets.kubernetes-client.io/volumes` is a JSON or
-YAML serizlied array of volume configuration objects:
+YAML serialized array of volume configuration objects:
 
 |Property|Type|Description|
 |--------|----|-----------|
--- a/api.md
+++ b/api.md
@@ -26,8 +26,8 @@ metadata:
   annotations:
     externalsecrets.kubernetes-client.io/volumes: >
       [
-        {"name": "db-secrets", "externalSecretName": "db-secrets"},
-        {"name": "client-secrets", "externalSecretName": "client-secrets"}
+        {"name": "db-secrets", "externalSecret": {"externalSecretName": "db-secrets"}},
+        {"name": "client-secrets", "externalSecret": {"externalSecretName": "client-secrets"}}
       ]
 spec:
   containers:
--- a/api.md
+++ b/api.md
@@ -30,7 +30,7 @@ should:
 * fetch the value of db/username from AWS Secrets Manager and write
   that value to the username file in the db-secrets volume;
 * create a key file in the client-secrets volume; and
-  fetch the value of api/key from AWS Secrets Manager and write that
+* fetch the value of api/key from AWS Secrets Manager and write that
   values to the api file in the client-secrets volume.
 
 ```yaml
--- a/api.md
+++ b/api.md
@@ -3,8 +3,8 @@
 Kubernetes External Secrets supports several "frontends" for storing
 secret data and presenting it to applications:
 
-* [Secret driver](#secret-frontend)
-* [Volume driver](#volume-frontend)
+* [Secret frontend](#secret-frontend)
+* [Volume frontend](#volume-frontend)
 
 ## Secret frontend
 
--- a/api.md
+++ b/api.md
@@ -13,10 +13,25 @@ TODO.
 ## Volume frontend
 
 The volume frontend writes secret data to volumes included in `Pod`
-specs. You configure which volumes the External Secret controller
-writes data to by adding the
+specs. The volume frontend implements a behavior analogous to the
+secret volume type. You specify `ExternalSecret` objects to use as
+volumes and the External Secret controller creates files containing
+secret data.
+
+To emulate an "externalSecret" volume type, you configure which
+volumes the External Secret controller writes data to by adding the
 `externalsecrets.kubernetes-client.io/volumes` annotation to a `Pod`
-manifest:
+manifest". With the manifest below, the External Secret controller
+should:
+
+* create password and username files in the db-secrets volume;
+* fetch the value of db/password from AWS Secrets Manager and write
+  that value to password file in the db-secrets volume;
+* fetch the value of db/username from AWS Secrets Manager and write
+  that value to the username file in the db-secrets volume;
+* create a key file in the client-secrets volume; and
+  fetch the value of api/key from AWS Secrets Manager and write that
+  values to the api file in the client-secrets volume.
 
 ```yaml
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


api.md
```diff
@@ -0,0 +1,63 @@
...
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  generateName: pod-example-
+  annotations:
+    externalsecrets.kubernetes-client.io/volumes: >
+      [
+        {"name": "db-secrets", "externalSecretName": "db-secrets"},
```
silasbw: Another way I was thinking about this was to mirror volume declarations.  E.g.,

```json
{"name": "db-secrets", "externalSecret": { "externalSecretName": "db-secrets" } }
```

https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod

I thought that my be too verbose.  On the other hand it parallels an existing Kubernetes pattern.

JacopoDaeli: Maybe 
```
{"name": "db-secrets", "externalSecret": { "name": "db-secrets" } }
```
?

jeffpearce: I like that even better than externalSecretName

JacopoDaeli: Yeah, it makes more sense.

silasbw: I also like "name" better than "externalSecretName", but it departs from the kubernetes convention (see the link I included above), so I'd encourage us to follow the convention of go with "externalSecretName".

silasbw:
@JacopoDaeli see https://github.com/godaddy/kubernetes-external-secrets/issues/46#issuecomment-498426782

silasbw:
@max-lobur do you have time to give this proposal a quick look?

api.md
```diff
@@ -0,0 +1,63 @@
...
+`externalsecrets.kubernetes-client.io/volumes` annotation to a `Pod`
+manifest:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  generateName: pod-example-
+  annotations:
+    externalsecrets.kubernetes-client.io/volumes: >
```
max-lobur: should be possible to write in yaml?
```
externalsecrets.kubernetes-client.io/volumes:
- name: db-secrets
  externalSecret:
    externalSecretName: db-secrets
...
```

max-lobur:
The last suggestion breaks the frontend abstraction tho

api.md
silasbw: Doh, yeah, updated.

silasbw:
Thanks @max-lobur !  Kubernetes External Secrets will create a file in the volume for each element in the `.secretDescriptor.data` array of the `ExternalSecret` specified in the annotation.  We're trying to minor the behavior of the native `Secret` object.  For example:

```yml
apiVersion: v1
kind: Pod
metadata:
  generateName: pod-example-
  annotations:
    externalsecrets.kubernetes-client.io/volumes: |
      - name: "hello-service"
        externalSecret:
          externalSecretName: "hello-service"
spec:
  containers:
  - image: busybox
    name: busybox
    volumeMounts:
    - mountPath: /secrets
      name: hello-service
  volumes:
  - name: hello-service
    emptyDir:
      medium: Memory
---
kind: ExternalSecret
metadata:
  name: hello-service
secretDescriptor:
  backendType: secretsManager
  data:
    - key: hello-service/password
      name: password
    - key: hello-service/apikey
      name: apikey
```

would cause the Kubernetes External Secrets controller to create two file in `/secrets`, one named `password` and one named `apikey`.  Kubernetes External Secrets would write the contents of `hello-service/password` from AWS SM into `password` and the contents of `hello-service/apikey` from AWS SM into `apikey`.

api.md
```diff
@@ -45,10 +62,32 @@ spec:
...
+  backendType: secretsManager
+  data:
+    - key: api/key
+      name: key
 ```
 
-The value of `externalsecrets.kubernetes-client.io/volumes` is a JSON
-array of volume configuration objects:
+The value of `externalsecrets.kubernetes-client.io/volumes` is a JSON or
+YAML serizlied array of volume configuration objects:
```
jeffpearce: serialized

api.md
```diff
@@ -0,0 +1,102 @@
...
+* fetch the value of api/key from AWS Secrets Manager and write that
+  values to the api file in the client-secrets volume.
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  generateName: pod-example-
+  annotations:
+    externalsecrets.kubernetes-client.io/volumes: |
```
max-lobur: Just curious, does the annotation has to be a multiline string `|`?
What if you just put a regular yaml there like
```
externalsecrets.kubernetes-client.io/volumes:
     - name: "db-secrets"
        externalSecret:
          externalSecretName: "db-secrets"
      - name: "client-secrets"
        externalSecret:
          externalSecretName: "client-secrets"
```
and then read it as array, without de-serealizing

silasbw: Kubernetes requires annotation values to be strings.

silasbw:
Applications could still leverage environment variables but you'd probably need to set some application code to do that instead of using `secretRef`.  *E.g.*, initContainer writes values to volume, containers source them before starting the application process (`source /some/volume; ./server.sh`).

Definitely open to better ways of doing this.  More ambitious would be adding a `volumeRef` to the API (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#envvarsource-v1-core)

There's definitely some trade-offs when moving off of etcd / `Secrets` to the volume-based frontend.  It's something folks have been asking for and in some cases believe it's a worthwhile improvement in privacy and integrity.

dudicoco:
Thanks for the reply @silasbw. 
I'm aware I can source environment variables beforehand, however, that wouldn't be very clean when used with third party applications since the Docker entrypoint will have to be replaced.

bank-vault has an interesting solution of injecting into the container's entrypoint:
https://banzaicloud.com/blog/inject-secrets-into-pods-vault-revisited/

"The init-container also changes the command of the container to run vault-env, instead of running the application directly"

Personally i'm leaning more towards using SSM along with native kubernetes secrets, but using the external-secrets controller means granting the pod access to all secrets (or permissions to assume role to all roles which can assume secrets) which is not ideal.

I think a good solution would be to bring up a job which uses the same IAM role as the deployment/statefulset/daemonset, the job will fetch the secret from Parameter Store and will update the appropriate secret in k8s.

Advantages:
1. Native k8s secrets
2. Granular access to Parameter Store.
3. One API call per deployment/statefulset/daemonset, less likely to throttle SSM API calls.
4. Can run as a cron job to keep secrets in sync with Parameter Store.

Let me know what you think!

silasbw:
Thanks for the bank-vault link. Not sure if that will work for `ExternalSecrets`or not, but it might help inspire a better solution.

I agree we should have more fine-grained roles on the controller.  That's something we should do regardless of how the secret data is communicated (`Secret` vs `Volume`) to the running containers.  FWIW, we are going to support the `Volume`-based approach as an alternative to the `Secret` one (not as a replacement).

For getting finer-grained roles, would it work well for you use case if we supported specifying a name space and a role for the controller to watch and assume?

dudicoco:
Do you have an estimation of when the volume based approach going to be implemented?

I think a nice alternative solution should give not take into account assuming a role, considering that it should be now possible to assign an IAM role directly to a pod's service account:
https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/

PR #96, opened 2019-06-21
fix: remove logging of potentially secret value

Flydiverny:
If one misconfigures a secret these can end up in logs, which could be quite a hassle :)

Example:

```yaml
...
secretDescriptor:
  backendType: secretsManager
  data:
      property: password
      name: password
```

hello-service/credentials from secrets manager:
```js
{
    "Name": "hello-service/credentials",
    "SecretString": "my-super-secret-password",
    ...
}
```

Would log the following 
```
Failed to JSON.parse 'my-super-secret-password': {}
```

### DIFF
--- a/lib/backends/kv-backend.js
+++ b/lib/backends/kv-backend.js
@@ -32,12 +32,14 @@ class KVBackend extends AbstractBackend {
         try {
           parsedValue = JSON.parse(value)
         } catch (err) {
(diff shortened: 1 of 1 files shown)


silasbw:
This change is included in `godaddy/kubernetes-external-secrets:1.3.0`.

PR #100, opened 2019-06-22
refactor: poller.js and daemon.js

silasbw:
Make it clear that there's a 1:1 mapping between a `Poller` and an
ExternalSecret.

Both files could use another refactor pass.


silasbw:
@Flydiverny, changes based on your feedback.

Flydiverny:
@silasbw looks good to me :)

PR #107, opened 2019-06-30
refactor: use watch api and instant poll new or modified secrets

Flydiverny:
Currently you can't set a lower events interval than poll interval, as this would result in never polling anything as the poller would be recreated everytime the events are polled.

Not sure if this is the best way to go about it but I tested refactoring so the events poller tracks which external secrets it has seen and yields appropriate events instead of always doing a delete all.

It would be nice if we could in some way instantly poll if an external secret has been modified or added. As if you have a long poll time you would currently have to wait for that to trigger, which feels slow when adding a new external secret. Not sure how to deal with pod restarts tho, as you wouldn't necessarily want to poll instantly when that happens, one alternative would be for the poller to check if the secret exists when started, and if it doesn't then trigger a poll right away.

Let me know what you think :) 

ping @silasbw

### DIFF
--- a/lib/external-secret.js
+++ b/lib/external-secret.js
@@ -2,6 +2,12 @@
 
 const JSONStream = require('json-stream')
 
+/**
+ * Creates an FIFO queue which you can put to and take from.
+ * If theres nothing to take it will wait with resolving until
+ * something is put to the queue.
+ * @returns {Object} Queue instance with put and take methods
+ */
 function createEventQueue () {
   const queuedEvents = []
   const waitingResolvers = []
--- a/lib/external-secret.js
+++ b/lib/external-secret.js
@@ -1,53 +1,88 @@
 'use strict'
 
-/**
- * Block asynchronous flow.
- * @param {number} milliseconds - Number of milliseconds to block flow operation.
- * @returns {Promise} Promise object representing block flow operation.
- */
-function sleep ({ milliseconds }) {
-  return new Promise(resolve => setTimeout(resolve, milliseconds))
+const JSONStream = require('json-stream')
+
+function createEventQueue () {
+  const queuedEvents = []
+  const waitingResolvers = []
+
+  return {
+    take: () => queuedEvents.length > 0
+      ? Promise.resolve(queuedEvents.shift())
+      : new Promise(resolve => waitingResolvers.push(resolve)),
+    put: (msg) => waitingResolvers.length > 0
+      ? waitingResolvers.shift()(msg)
+      : queuedEvents.push(msg)
+  }
+}
+
+async function startWatcher ({
+  kubeClient,
+  customResourceManifest,
+  logger,
+  eventQueue
+}) {
+  const deathQueue = createEventQueue()
+
+  try {
+    while (true) {
+      logger.info('Starting watch stream')
+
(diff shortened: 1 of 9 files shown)


silasbw:
@Flydiverny an alternative to polling is using the Kubernetes API watch endpoints. I think we want to use those eventually, but hadn't gotten around to the implementation. If we're going to make improvements to this code, it would be good for those improvements to work with the watch implementation (or maybe we do that one first?).

@satish-ravi, thoughts?

Flydiverny:
yeah I was wondering why it didn't use watch, but still tried to simulate it somehow? :D So I went with yielding event types based on what watch should provide, not sure about the payload tho. I'm not sure how to use watch with the kubernetes-client though 

would you do something like?
```js
    const stream = kubeClient
          .apis[customResourceManifest.spec.group]
          .v1[customResourceManifest.spec.names.plural]
          .get({ watch: true })
    const jsonStream = new JSONStream()
    stream.pipe(jsonStream)
    jsonStream.on('data', object => {
      // do stuff
    })
```

Flydiverny:
@silasbw made some updates, now using the watch api instead.

jeffpearce:
Hi @Flydiverny. I'm taking a look at this today

jeffpearce:
@satish-ravi

Flydiverny:
Yeah, I'll update the test to work with the stream.
I'm not sure about how to deal with error handling tho, what happens when it loses connection, how to get it to recover?
Think this makes it harder as well https://github.com/godaddy/kubernetes-client/issues/248?  kubernetes-client is also deprecating `getStream` but there's no `getObjectStream` for the CRDs.

lib/external-secret.js
```diff
@@ -1,53 +1,64 @@
...
+
+    jsonStream.on('data', eventQueue.put)
+
+    jsonStream.on('error', (err) => {
+      logger.warn(err, 'Got error on stream')
+      eventQueue.put({ type: 'DELETE_ALL' })
+    })
+
+    jsonStream.on('end', () => {
+      logger.warn('Watch stream ended')
```
satish-ravi: This could potentially make the logs noisy. The kube-apiserver disconnects streams periodically (see min-request-timeout here https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/)

silasbw:
Yes, we should add `getObjectStream`.

Flydiverny:
Please check again if there is something more to be done, or that you want reverted :)

jeffpearce:
Hi @Flydiverny sorry for the delay. I'll take another look today.

lib/external-secret.js
```diff
@@ -1,53 +1,94 @@
...
+  kubeClient,
+  customResourceManifest,
+  logger,
+  eventQueue
+}) {
+  const deathQueue = createEventQueue()
+
+  try {
+    while (true) {
+      logger.info('Starting watch stream')
```
satish-ravi: To be consistent with stop log
```suggestion
      logger.debug('Starting watch stream')
```

Flydiverny:
On vacation so can't follow up on the failed build for some days  not sure why it decided to fail suddenly :)

Issue #92, opened 2019-06-10
Will this work on google container engine kubernetes?


red8888:
On EKS I can just use roles but if Im on GKE I cant to that

Can I use the default AWS environment variables to set the secret access key and id so I can create an IAM user my gke clusters can use to get access to secrets manager?

silasbw:
Yep, that sounds like it should work, but I haven't heard of anyone trying that yet.  If you try it out and it works let us know or send us a PR to update the README.md.

ajaysheoran2323:
@red8888  Are you able to use this process for GKE?

red8888:
haven't tested yet but will soon. I mean its a pretty quick test, just set the AWS key and secret key env vars. you could probably just pull the container and run locally to test this I suspect

It looks like this is intended to run on EKS with AWS roles which means they are probably instantiating the clients without specifying creds/roles. If thats the case then this should totally work- all the AWS APIs look for those creds by default

red8888:
Tried this but does not appear to be working I am seeing this int he logs:
"failure while polling the secrets {}"   
Im not seeing what the actual error is beyond that message

I confirmed these env vars are set in the pod and the creds have access to get my secrets:
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY

red8888:
ok @silasbw I just cloned this repo and to add some additional logging messages and pushed the image to gcr to test

when deploying with that image Its totally working now- I made no code changes other than adding some log messages

Was this fixed and there is a new release waiting to be cut or something? before I was running the latest image (godaddy/kubernetes-external-secrets:latest)

red8888:
OK! im dumb your "latest" tag is 4 months old. I pulled the actual latest tag "1.3.0" and now its working yay!

Can I send a PR or could someone just up the chart with support for env vars from secrets? I wouldn't hard code params for aws keys it could easily be more dynamic and just support hydrating env vars from _any_ k8s secret

I have been doing this in my charts which works really well for me:

values.yaml:
```
  # define env vars to be created from existing k8s secrets
  envVarsFromSecret:
    AWS_ACCESS_KEY_ID:
      secretKeyRef: aws-credentials
      key: id
    AWS_SECRET_ACCESS_KEY:
      secretKeyRef: aws-credentials
      key: key
```

Then in the deployment:
```
    .......
    spec:
      serviceAccountName: {{ template "kubernetes-external-secrets.fullname" . }}
      containers:
[...]

PR #108, opened 2019-07-01
Support env vars from secrets in chart

red8888:
per this issue I opened: https://github.com/godaddy/kubernetes-external-secrets/issues/92

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,23 +50,25 @@ helm delete kubernetes-external-secrets
 
 The following table lists the configurable parameters of the `kubernetes-external-secrets` chart and their default values.
 
-| Parameter                            | Description                                                  | Default                                                 |
-| ------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------- |
-| `env.AWS_REGION`                     | Set AWS_REGION in Deployment Pod                             | `us-west-2`                                             |
-| `env.EVENTS_INTERVAL_MILLISECONDS`   | Set EVENTS_INTERVAL_MILLISECONDS in Deployment Pod           | `60000`                                                 |
-| `env.POLLER_INTERVAL_MILLISECONDS`   | Set POLLER_INTERVAL_MILLISECONDS in Deployment Pod           | `10000`                                                 |
-| `image.repository`                   | kubernetes-external-secrets Image name                       | `godaddy/kubernetes-external-secrets`                   |
-| `image.tag`                          | kubernetes-external-secrets Image tag                        | `1.2.0`                                                 |
-| `image.pullPolicy`                   | Image pull policy                                            | `IfNotPresent`                                          |
-| `rbac.create`                        | Create & use RBAC resources                                  | `true`                                                  |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -50,23 +50,25 @@ helm delete kubernetes-external-secrets
...
+| Parameter                                 | Description                                                  | Default                                                 |
+| ----------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------- |
+| `env.AWS_REGION`                          | Set AWS_REGION in Deployment Pod                             | `us-west-2`                                             |
+| `env.EVENTS_INTERVAL_MILLISECONDS`        | Set EVENTS_INTERVAL_MILLISECONDS in Deployment Pod           | `60000`                                                 |
+| `env.POLLER_INTERVAL_MILLISECONDS`        | Set POLLER_INTERVAL_MILLISECONDS in Deployment Pod           | `10000`                                                 |
+| `envVarsFromSecret.AWS_ACCESS_KEY_ID`     | Set AWS_ACCESS_KEY_ID (from a secret) in Deployment Pod      |                                                         |
+| `envVarsFromSecret.AWS_SECRET_ACCESS_KEY` | Set AWS_SECRET_ACCESS_KEY (from a secret) in Deployment Pod  |                                                         |
```
jeffpearce: Curious why `envVarsFromSecret` instead of `env`? The values can be provided via the helm chart, but they can also be provided by other ways (e.g. `export`). Is there a way to express both cases here?

Flydiverny:
Wouldn't it be enough to just add templating for

```yaml
{{- if .Values.aws.secret.name }}
          env:
            - name: AWS_ACCESS_KEY_ID
              valueFrom:
                secretKeyRef:
                  key: aws_access_key_id
                  name: {{ .Values.aws.secret.name }}
            - name: AWS_SECRET_ACCESS_KEY
              valueFrom:
                secretKeyRef:
                  key: aws_secret_key
                  name: {{ .Values.aws.secret.name }}
{{- end }}
```

and recommend users to create a secret and provide the name :)

```sh
$ kubectl create secret generic <SECRET_NAME> --from-literal=aws_access_key_id=$AWS_ACCESS_KEY_ID --from-literal=aws_secret_key=$AWS_SECRET_ACCESS_KEY
$ helm install --name kubernetes-external-secrets --set aws.secret.name=<SECRET_NAME>
```

jeffpearce:
That works for me

red8888:
@flydiverny I avoided hard coding the AWS creds use case- I've been really frustrated with overly specific helm charts I see in the wild. If it supports creating _any_ env vars from secrets you won't have to keep extending it for additional use cases and have this arbitrary limit on it's configurability

README.md
red8888: envVarsFromSecret is prob not a great name. Were you thinking maybe nesting it under env? I'm okay with that. I'm new to helm and I keep reading about how you should keep your configs as "flat" as possible- but new to helm and don't have much reference to draw from. Or did you just mean make the documentation clearer?

Issue #89, opened 2019-06-06
No way to install via kubectl?


cwiggs:
I was directed here from this blog: https://godaddy.github.io/2019/04/16/kubernetes-external-secrets/

I was looking into using this tool but noticed that recently deploying via kubectl was removed and helm is now the official way to install.  Is there some type of workaround or plans to re-introduce deploying with kubectl?  Having to install helm just to install this tool is a real bummer.

PR where helm was setup as the official way to install: #68 
When external-secrets.yml was removed: #70 
Someone else asked a similar question here: #72

silasbw:
Yes, we've been planning on re-adding this!

Thanks for creating the issue to track.  We'd welcome a PR too :)

cwiggs:
Can you give me more insight into what needs to be done or why the kubectl option was removed?  I have a k8 cluster I can test with and am comfortable with debugging stuff.  Not sure if I have time to add a ton to the code base though.

silasbw:
I'll merge https://github.com/godaddy/kubernetes-external-secrets/pull/90, but we need a few improvements:

- [ ] Update README.md with instructions
- [ ] Decide how we want apply runtime configuration (_e.g._, AWS_REGION)

For the runtime config, we could add a ConfigMap to the external-secrets.yml file that has a default AWS_REGION; and update the README.md with instructions on how to update it and restart the controller.

Does that seem reasonable @cwiggs ?

Flydiverny:
Should we close this with the merge of #212 ?
cc @keweilu @silasbw

PR #118, opened 2019-07-15
deploy via kubectl

jeffpearce:
Add instructions and examples for deploying external secrets manager via kubectl. Addresses #89

### DIFF
--- a/deploy/deploy.yml
+++ b/deploy/deploy.yml
@@ -1,0 +1,50 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: kubernetes-external-secrets
+  labels:
+    app.kubernetes.io/name: kubernetes-external-secrets
+
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: kubernetes-external-secrets
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: kubernetes-external-secrets
+    spec:
+      containers:
+        - name: kubernetes-external-secrets
+          image: godaddy/kubernetes-external-secrets:1.3.0
+          envFrom:
+          - configMapRef:
+              name: external-secrets-config
+          # env:
+          #   - name: AWS_REGION
+          #     valueFrom:
+          #       configMapKeyRef:
+          #         name: external-secrets-config
+          #         key: AWS_REGION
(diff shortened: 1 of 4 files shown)


jeffpearce:
@cwiggs, if you have time, could you let me know whether this addresses your needs?

deploy/external-secrets-env.properties
```diff
@@ -0,0 +1,9 @@
+# kubectl -n=es-kubectl create configmap external-secrets-config --from-env-file=deploy/external-secrets-env.properties
```
silasbw: Should we commit the `ConfigMap` instead of the env file that we then create a `ConfigMap` with?  I'd also include the `ConfigMag` and `Deployment` in one yaml file.

deploy/deploy.yml
```diff
@@ -0,0 +1,50 @@
+apiVersion: extensions/v1beta1
```
silasbw: Do we need the RBAC resources?

jeffpearce: I thought rbac was an option? I can provide a sample yaml to create them - might be better to have that in a separate file though

deploy/external-secrets-env.properties
jeffpearce: I don't feel strongly about putting them in one file, but one advantage of them being separate is that it makes it easier to update the configmap separately if you need to change it

jeffpearce: Actually, since it's a sample, we could provide both

deploy/deploy.yml
silasbw: This example is unlikely to work on production clusters because they're probably using RBAC.  I'd include it.

deploy/deploy.yml
```diff
@@ -0,0 +1,50 @@
...
+      labels:
+        app.kubernetes.io/name: kubernetes-external-secrets
+    spec:
+      containers:
+        - name: kubernetes-external-secrets
+          image: godaddy/kubernetes-external-secrets:1.3.0
+          envFrom:
+          - configMapRef:
+              name: external-secrets-config
+          # env:
```
silasbw: Uncomment this and include the ConfigMap?

jeffpearce: I have it locally - will clean up and push the changes

silasbw:
Closing due to inactivity.  Happy to resurrect it if we should.

PR #131, opened 2019-07-23
Revert "fix(daemon): Upsert secrets immediately poller added "

Flydiverny:
This reverts commit a986dfba1c96679faab0097b96c5f77e4deb8dca.

I'd like to revert the change introduced in #125 as it causes unnecessary polling when watch stream is disconnected, and causes upserts for all external secrets everytime the pollers are recreated.

I think we can definitely do some kind of improvement here tho.

See discussion in #117 :)

### DIFF
--- a/lib/daemon.js
+++ b/lib/daemon.js
@@ -96,7 +96,7 @@ class Daemon {
         }
 
         case 'ADDED': {
-          this._addPoller(descriptor, true)
+          this._addPoller(descriptor)
           break
         }
 


jeffpearce:
Hi @Flydiverny - see my comment in other PR. Can you help me understand why the extra polling is bad?

For my .02 I prefer the simpler code, but I'm open to being convinced otherwise  :-)

Flydiverny:
Closing this, see #133

PR #133, opened 2019-07-23
feat: add status subresource with last sync and generation tracking

### LINKED ISSUE #117: help: force secrets upserting eventhough POLLER_INTERVAL_MILLISECONDS is set to higher interval

### LINKED ISSUE #153: high poller interval problem
Hi, I noticed if I set POLLER_INTERVAL_MILLISECONDS=300000 or to a bigger value,
it will not create any secrets even after an hour and in logs I see successful status
```{"level":30,"time":1566397148667,"pid":17,"hostname":"external-secrets-kubernetes-external-secrets-7d65747f56-5h5lw","msg":"starting poller","v":1}
{"level":30,"time":1566397148667,"pid":17,"hostname":"external-secrets-kubernetes-external-secrets-7d65747f56-5h5lw","msg":"spinning up poller {\"id\":\"7815d6df-c41c-11e9-a8dd-0622d2c58d44\",\"namespace\":\"default\",\"secretDescriptor\":{\"backendType\":\"secretsManager\",\"data\":[{\"key\":\"hello-service/password\",\"name\":\"password\"}],\"name\":\"hello-service2\"},\"ownerReference\":{\"apiVersion\":\"kubernetes-client.io/v1\",\"controller\":true,\"kind\":\"ExternalSecret\",\"name\":\"hello-service2\",\"uid\":\"7815d6df-c41c-11e9-a8dd-0622d2c58d44\"}}","v":1}
{"level":30,"time":1566397148668,"pid":17,"hostname":"external-secrets-kubernetes-external-secrets-7d65747f56-5h5lw","msg":"starting poller","v":1}
```
app version 1.3.1

### LINKED ISSUE #190: Periodically ADDED
Even the External Secrets was created once, but external-secrets controller get ADDED event periodically. It makes unnecessary SSM API calling because of the CRD is not changed. 
```
{"level":30,"time":1572413517121,"pid":16,"hostname":"kubernetes-external-secrets-574cb46f86-rd68v","msg":"externalSecretEvents: ADDED","v":1}
{"level":30,"time":1572413605575,"pid":16,"hostname":"kubernetes-external-secrets-574cb46f86-rd68v","msg":"externalSecretEvents: DELETED_ALL","v":1}
{"level":30,"time":1572413605607,"pid":16,"hostname":"kubernetes-external-secrets-574cb46f86-rd68v","msg":"externalSecretEvents: ADDED","v":1}
{"level":30,"time":1572413715790,"pid":16,"hostname":"kubernetes-external-secrets-574cb46f86-rd68v","msg":"externalSecretEvents: DELETED_ALL","v":1}
{"level":30,"time":1572413715803,"pid":16,"hostname":"kubernetes-external-secrets-574cb46f86-rd68v","msg":"externalSecretEvents: ADDED","v":1}
{"level":30,"time":1572413805861,"pid":16,"hostname":"kubernetes-external-secrets-574cb46f86-rd68v","msg":"externalSecretEvents: DELETED_ALL","v":1}
{"level":30,"time":1572413805866,"pid":16,"hostname":"kubernetes-external-secrets-574cb46f86-rd68v","msg":"externalSecretEvents: ADDED","v":1}
```

How do I handle this?

Flydiverny:
Currently kubernetes-external-secrets does excessive polling of all external secrets and doesn't respect the polling intervall. Everytime the watch stream disconnects (happens somewhat regularly) all the secrets are regenerated. In our relatively small dev cluster this generates 22 million events to each of KMS, SM and GuardDuty which generates unnecessary billing, and a general waste of compute cycles for the sake of nothing.

This PR uses the status subresource to track generation of external secret, and the last sync time.
Will use these values to determine how to queue up the next poll, either instant or time remaining for next poll intervall to pass. This helps for setting long poll intervals as currently setting something like 24h and having the pod moved or watch stream disconnected would restart the interval from that time, instead of forcing unwanted polls.

Updated CRD to show fields when using kubectl:
image

closes #131 
fixes #117
fixes #153 
fixes #190
start of #185

### DIFF
--- a/charts/kubernetes-external-secrets/templates/rbac.yaml
+++ b/charts/kubernetes-external-secrets/templates/rbac.yaml
@@ -25,6 +25,9 @@ rules:
   - apiGroups: ["kubernetes-client.io"]
     resources: ["externalsecrets"]
     verbs: ["get", "watch", "list"]
+  - apiGroups: ["kubernetes-client.io"]
+    resources: ["externalsecrets/status"]
+    verbs: ["get", "update"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRoleBinding
--- a/lib/daemon.js
+++ b/lib/daemon.js
@@ -52,10 +52,6 @@ class Daemon {
   _addPoller (descriptor) {
     this._logger.info('spinning up poller for', descriptor.name, 'in', descriptor.namespace)
 
-    if (this._pollers[descriptor.id]) {
-      throw new Error(`Poller for ${descriptor.id} already exists`)
-    }
-
     const poller = this._pollerFactory.createPoller(descriptor)
 
     this._pollers[descriptor.id] = poller.start()
@@ -74,11 +70,7 @@ class Daemon {
           break
         }
 
-        case 'ADDED': {
-          this._addPoller(descriptor)
-          break
-        }
-
+        case 'ADDED':
         case 'MODIFIED': {
           this._removePoller(descriptor.id)
           this._addPoller(descriptor)
--- a/lib/daemon.js
+++ b/lib/daemon.js
@@ -1,9 +1,5 @@
 'use strict'
 
-/* eslint-disable no-console */
-
-const Poller = require('./poller')
-
 /** Daemon class. */
 class Daemon {
   /**
@@ -15,21 +11,13 @@ class Daemon {
    * @param {number} pollerIntervalMilliseconds - Interval time in milliseconds for polling secret properties.
    */
   constructor ({
-    backends,
     externalSecretEvents,
-    kubeClient,
     logger,
-    metrics,
-    pollerIntervalMilliseconds,
-    customResourceManifest
+    pollerFactory
   }) {
-    this._backends = backends
-    this._kubeClient = kubeClient
     this._externalSecretEvents = externalSecretEvents
     this._logger = logger
-    this._metrics = metrics
-    this._pollerIntervalMilliseconds = pollerIntervalMilliseconds
-    this._customResourceManifest = customResourceManifest
+    this._pollerFactory = pollerFactory
 
     this._pollers = {}
   }
@@ -68,15 +56,7 @@ class Daemon {
       throw new Error(`Poller for ${descriptor.id} already exists`)
     }
 
-    const poller = new Poller({
-      backends: this._backends,
-      intervalMilliseconds: this._pollerIntervalMilliseconds,
-      kubeClient: this._kubeClient,
-      logger: this._logger,
-      metrics: this._metrics,
-      customResourceManifest: this._customResourceManifest,
-      externalSecret: descriptor.externalSecret
-    })
+    const poller = this._pollerFactory.createPoller(descriptor)
 
     this._pollers[descriptor.id] = poller.start()
   }
--- a/lib/daemon.js
+++ b/lib/daemon.js
@@ -20,7 +20,8 @@ class Daemon {
     kubeClient,
     logger,
     metrics,
-    pollerIntervalMilliseconds
+    pollerIntervalMilliseconds,
+    customResourceManifest
   }) {
     this._backends = backends
     this._kubeClient = kubeClient
@@ -28,6 +29,7 @@ class Daemon {
     this._logger = logger
     this._metrics = metrics
     this._pollerIntervalMilliseconds = pollerIntervalMilliseconds
+    this._customResourceManifest = customResourceManifest
 
     this._pollers = {}
   }
@@ -39,17 +41,8 @@ class Daemon {
    */
   _createPollerDescriptor (externalSecret) {
     const { uid, name, namespace } = externalSecret.metadata
-    // NOTE(jdaeli): hash this in case resource version becomes too long?
-    const secretDescriptor = { ...externalSecret.secretDescriptor, name }
-    const ownerReference = {
-      apiVersion: externalSecret.apiVersion,
-      controller: true,
-      kind: externalSecret.kind,
-      name,
-      uid
-    }
 
-    return { id: uid, namespace, secretDescriptor, ownerReference }
+    return { id: uid, name, namespace, externalSecret }
   }
 
   /**
@@ -68,8 +61,12 @@ class Daemon {
     Object.keys(this._pollers).forEach(pollerId => this._removePoller(pollerId))
   }
 
-  _addPoller (descriptor, forcePoll = false) {
-    this._logger.info('spinning up poller', descriptor)
+  _addPoller (descriptor) {
+    this._logger.info('spinning up poller for', descriptor.name, 'in', descriptor.namespace)
+
+    if (this._pollers[descriptor.id]) {
+      throw new Error(`Poller for ${descriptor.id} already exists`)
+    }
 
     const poller = new Poller({
       backends: this._backends,
@@ -77,12 +74,11 @@ class Daemon {
       kubeClient: this._kubeClient,
       logger: this._logger,
       metrics: this._metrics,
-      namespace: descriptor.namespace,
-      secretDescriptor: descriptor.secretDescriptor,
-      ownerReference: descriptor.ownerReference
+      customResourceManifest: this._customResourceManifest,
+      externalSecret: descriptor.externalSecret
     })
 
-    this._pollers[descriptor.id] = poller.start({ forcePoll })
+    this._pollers[descriptor.id] = poller.start()
   }
 
   /**
@@ -99,13 +95,13 @@ class Daemon {
         }
 
         case 'ADDED': {
-          this._addPoller(descriptor, true)
+          this._addPoller(descriptor)
           break
         }
 
         case 'MODIFIED': {
           this._removePoller(descriptor.id)
-          this._addPoller(descriptor, true)
+          this._addPoller(descriptor)
           break
         }
 
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -99,6 +99,13 @@ class Poller {
     try {
       await this._upsertKubernetesSecret()
       await this._updateStatus('SUCCESS')
+
+      this._metrics.observeSync({
+        name: this._name,
+        namespace: this._namespace,
+        backend: this._secretDescriptor.backendType,
+        status: 'success'
+      })
     } catch (err) {
       this._logger.error(err, `failure while polling the secret ${this._name}`)
       await this._updateStatus(`ERROR, ${err.message}`)
@@ -110,13 +117,6 @@ class Poller {
         status: 'error'
       })
     }
-
-    this._metrics.observeSync({
-      name: this._name,
-      namespace: this._namespace,
-      backend: this._secretDescriptor.backendType,
-      status: 'success'
-    })
   }
 
   /**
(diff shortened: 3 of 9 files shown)


Flydiverny:
@silasbw @jeffpearce Please have a look :D

jeffpearce:
Hi @Flydiverny. It seems like it's much simpler to just always upsert the secret when we start up. Do you have concerns around doing it that way?

Flydiverny:
@jeffpearce Cost would be a concern, and also it defeats the purpose of having a polling interval setting if its not respected.

As mentioned in #117 having a longer interval (which we do as well) the interval will currently (version 1.3.1) never happen or possibly happen in a totally different interval if either: the watch stream disconnects or the pod is restarted.

Say the watch stream doesn't seem to remain connected forever so if I then use a 24h interval and the watch stream only lasts 1h, I would never trigger the interval at 24h because every hour the 24h countdown is restarted as the poller restarts. I don't see any way around this without keeping the state somewhere (which in this PR is in the created secrets).

Further building onto this I think we could allow for specifying a polling interval per external secret as well. Could be useful for testing or in cases where some secrets rotate more often. Possibly even disabling polling for some and only update when external secret is updated.

silasbw:
Hi @Flydiverny, which cost are you thinking of?  The cost of polling AWS SM?

(Sorry for the delay, been on vacation)

chadlwilson:
Any chance of us getting a new release with better support for the longer poll intervals either with/via this PR or with the simpler fix already on master?

Flydiverny:
> Hi @Flydiverny, which cost are you thinking of? The cost of polling AWS SM?
> 
> (Sorry for the delay, been on vacation)

Yes this was referencing the costs of polling AWS SM, after running external-secrets for awhile now in our environments we can see we get unnecessarily expensive bills for the excessive amount of polling that is done.

Flydiverny:
Refactored to use status subresource instead.
Added generation tracking and updated CRD to show last sync and a status when using kubectl
image

Got some tests to fix if there's any interest in this.

Refactor will also got a bit of the way on #185 

See also #190 which points out the issue I'm trying to solve here :)

Flydiverny:
Tests fixed and stuff squashed and rebased to madness.

Flydiverny:
For anyone that wants to try this i pushed an image here https://hub.docker.com/r/flydiverny/kubernetes-external-secrets/tags

```sh
docker pull flydiverny/kubernetes-external-secrets:1.6.0-with.pr.133
```

Flydiverny:
@jeffpearce @silasbw @keweilu Please see updated description and review updated PR! :)

chadlwilson:
Thanks for your work here @Flydiverny 🎉

lib/poller.js
```diff
@@ -74,49 +94,68 @@ class Poller {
...
     try {
-      return await kubeNamespace.secrets.post({ body: secretManifest })
+      await kubeNamespace.secrets.post({ body: secretManifest })
     } catch (err) {
       if (err.statusCode !== 409) throw err
-      return kubeNamespace.secrets(this._secretDescriptor.name).put({ body: secretManifest })
+      await kubeNamespace.secrets(this._name).put({ body: secretManifest })
     }
+
+    await this._updateStatus('SUCCESS')
```
silasbw: nit: move this to `._updateStatus` to https://github.com/godaddy/kubernetes-external-secrets/pull/133/files#diff-ef923d397aa8d0a9b6bc6d401b5290d7R101.  Then the 'SUCCESS' and 'ERROR' status setting are in the same function. When I was reading this the first time I aw the 'ERROR' status and wondered why there wasn't a 'SUCCESS' status near by.

lib/poller.js
```diff
@@ -150,37 +189,66 @@ class Poller {
...
+      const currentGeneration = this._externalSecret.metadata.generation
+
+      if (observedGeneration < currentGeneration) {
+        return this._setNextPoll(0)
+      }
+
+      const lastPollNbr = Date.parse(lastSync) || 0
+      const now = Date.now()
+
+      if (lastPollNbr > now) {
```
silasbw: In what cases would you expect this to be true?

lib/poller.js
```diff
@@ -150,37 +189,66 @@ class Poller {
...
+        } = {}
+      } = await this._status.get()
+
+      const currentGeneration = this._externalSecret.metadata.generation
+
+      if (observedGeneration < currentGeneration) {
+        return this._setNextPoll(0)
+      }
+
+      const lastPollNbr = Date.parse(lastSync) || 0
```
silasbw: I don't understand this variable name.  What's `Nbr`?

How about `lastPollDate` or `lastPollTime`?

lib/poller.js
```diff
@@ -150,37 +189,66 @@ class Poller {
...
+      const now = Date.now()
+
+      if (lastPollNbr > now) {
+        return this._setNextPoll(this._intervalMilliseconds)
       }
+
+      const nextPollIn = Math.max(lastPollNbr - (now - this._intervalMilliseconds), 0)
```
silasbw: Can you add a variable to make this slightly easier to read?

```js
const elapsedTime = now - lastPollTime
const nextPollIn = Math.max(this._intervalMilliseconds - elaspedTime, 0)
```

lib/poller.js
```diff
@@ -150,37 +189,66 @@ class Poller {
...
   /**
-   * Checks if secret exists, if not trigger a poll
+   * Checks status of external secret and determines time to next poll
+   * If current observed generation is older than ES generation it will poll right away
+   * otherwise check when it was last polled and set timeout for next poll
    */
-  async _checkForSecret () {
-    const kubeNamespace = this._kubeClient.api.v1.namespaces(this._namespace)
-
+  async _checkExternalSecretStatus () {
```
silasbw: naming nit: how about `_scheduleNextPoll`?  I think that more accurately describes what this function does.

lib/poller.js
```diff
@@ -23,27 +23,47 @@ class Poller {
    * @param {Object} kubeClient - Client for interacting with kubernetes cluster.
    * @param {Object} logger - Logger for logging stuff.
    * @param {string} namespace - Kubernetes namespace.
-   * @param {SecretDescriptor} secretDescriptor - Kubernetes secret descriptor.
+   * @param {Object} customResourceManifest - CRD manifest
+   * @param {Object} externalSecret - ExternalSecret custom resource object.
```
silasbw: comment nit: "ExternalSecret manifest" instead of "ExternalSecret custom resource object".  When I see "custom resource object" i think CRD / Custom Resource Definition object.

lib/daemon.js
```diff
@@ -68,21 +49,16 @@ class Daemon {
...
-
-    this._pollers[descriptor.id] = poller.start({ forcePoll })
+  _addPoller (descriptor) {
+    this._logger.info('spinning up poller for', descriptor.name, 'in', descriptor.namespace)
+
+    if (this._pollers[descriptor.id]) {
+      throw new Error(`Poller for ${descriptor.id} already exists`)
+    }
+
+    const poller = this._pollerFactory.createPoller(descriptor)
```
silasbw: I'm a little confused at the types here: `descriptor` is result of `_createPollerDescriptor`, but `.createPoller` takes an `ExternalSecret` manifest / object.  Is there an actual type problem, or confusion with variable naming?

lib/poller.js
Flydiverny: Thanks it didn't feel right :smile:

lib/daemon.js
Flydiverny: Hmm well the full externalSecret now exists in in the descriptor, just destructuring externalSecret from the descriptor as nothing more is needed to create the poller, any suggestions to make it more clear?

Ie `this._pollerFactory.createPoller({ externalSecret: descriptor.externalSecret} )`

lib/poller.js
Flydiverny: Well never TBH, unless someone messes up their time zones or so :D I just ran into it when messing with the tests (intentionally setting bad dates). Could remove it

lib/poller.js
Flydiverny: Bad naming from first pass :tada: I'll rename it to time since it's ms

lib/daemon.js
```diff
@@ -68,21 +49,16 @@ class Daemon {
...
-      ownerReference: descriptor.ownerReference
-    })
-
-    this._pollers[descriptor.id] = poller.start({ forcePoll })
+  _addPoller (descriptor) {
+    this._logger.info('spinning up poller for', descriptor.name, 'in', descriptor.namespace)
+
+    if (this._pollers[descriptor.id]) {
+      throw new Error(`Poller for ${descriptor.id} already exists`)
+    }
```
Flydiverny: So after running this for a few hours, this seems to happen occasionally, not really sure whats triggering it. 

Whenever an external secret is polled we will update the status for this external secret, this in turn should trigger the watch stream with a modified event. Which means we recreate the poller and schedule the next poll. As we track the generation with observedGeneration we won't end up in an endless loop as generation is not updated by status updates.

Flydiverny: With change in bb3787c this now seems fine.

Switched back to returning in _upsertSecret here: https://github.com/godaddy/kubernetes-external-secrets/pull/133/commits/cf17824cde407c9bfed571b4bd72c20b1e9a3fe9
But this seemed to cause issues with errors getting caught in the wrong try catch. ie upserts failing. so added the await again https://github.com/godaddy/kubernetes-external-secrets/pull/133/commits/c8be7f389390a9e7de8046b2d28607f39c533a5a

Running test in our dev cluster again 🎉

lib/daemon.js
silasbw: doh, i missed the destructuring in the `.createPoller`.  ignore me :)

lib/poller.js
```diff
@@ -150,37 +189,62 @@ class Poller {
...
+      } = await this._status.get()
+
+      const currentGeneration = this._externalSecret.metadata.generation
+
+      if (observedGeneration < currentGeneration) {
+        return this._setNextPoll(0)
       }
+
+      const lastPollTime = Date.parse(lastSync) || 0
+      const elapsedTime = Date.now() - lastPollTime
```
silasbw: I didn't mean to imply we should remove it.  I like the defensive behavior of scheduling it immediately (and then hopefully getting a more sane lastPollTime value).  Maybe bring the code back and add a comment?

Flydiverny: Ah yes, must have been tired last night. I was thinking this became a non issue after the changes but a negative elapsedTime here gives a funny next poll time 😅 I'll add it back.

charts/kubernetes-external-secrets/templates/rbac.yaml
```diff
@@ -25,6 +25,9 @@ rules:
   - apiGroups: ["kubernetes-client.io"]
     resources: ["externalsecrets"]
     verbs: ["get", "watch", "list"]
+  - apiGroups: ["kubernetes-client.io"]
```
keweilu: can we also update the rbac in external-secrets.yml file https://github.com/godaddy/kubernetes-external-secrets/blob/master/external-secrets.yml#L18 with those permissions

Flydiverny: Fixt !

Issue #137, opened 2019-07-27
Support for Annotations


cabrinha:
# Support for Annotations

Thanks for the great piece of software. I'm looking to use external-secrets alongside [kubernetes-replicator](https://github.com/mittwald/kubernetes-replicator) and in order to automatically duplicate my external secrets to different namespaces, they need to contain an annotation of: 

```yaml
    replicator.v1.mittwald.de/replication-allowed: "true"
    replicator.v1.mittwald.de/replication-allowed-namespaces: .*
```

However, when I create my external secret with these annotations, they are not carried over into the kubernetes secret object, preventing them from being replicated.

Please consider passing all/some annotations from the ExternalSecret object to the Secret object.

## Steps to reproduce

Create an ExternalSecret with the annotations:
```yaml
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  annotations:
    replicator.v1.mittwald.de/replication-allowed: "true"
    replicator.v1.mittwald.de/replication-allowed-namespaces: .*
  name: datadog
  namespace: default
secretDescriptor:
  backendType: secretsManager
  data:
  - key: datadog
    name: datadog-api-key
    property: datadog-api-key
  - key: datadog
    name: datadog-auth-token
    property: datadog-auth-token
```

See that the resulting Secret does not contain annotations:

```yaml
apiVersion: v1
data:
  datadog-api-key: xxx
  datadog-auth-token: xxx
kind: Secret
metadata:
  creationTimestamp: "2019-07-28T16:26:16Z"
  name: datadog
  namespace: default
  ownerReferences:
  - apiVersion: kubernetes-client.io/v1
    controller: true
    kind: ExternalSecret
    name: datadog
    uid: 86859120-affc-11e9-a909-025f2055801a
  resourceVersion: "649857"
  selfLink: /api/v1/namespaces/default/secrets/datadog
  uid: 6c3f8ecf-b154-11e9-a909-025f2055801a
type: Opaque
```

Flydiverny:
Setting secret annotations are now supported through `template`

```yaml
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  name: test
spec:
  backendType: systemManager
  template:
    metadata:
      annotations:
        my-annotation: "hello world"# <--
  data:
  - key: /test/test
    name: test-test
```

which should give you a secret like:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test
  annotations:
[...]

PR #139, opened 2019-07-29
feat: adding annotations support

cabrinha:
Resolves https://github.com/godaddy/kubernetes-external-secrets/issues/137 by allowing annotations to be passed through the ExternalSecret and into the resulting Secret.

Not quite sure how to get the tests passing, but I've tested this on my cluster and it works with and without annotations added to the ExternalSecret object.

### DIFF
--- a/lib/daemon.js
+++ b/lib/daemon.js
@@ -36,7 +36,7 @@ class Daemon {
    * @returns {Object} Poller descriptor.
    */
   _createPollerDescriptor (externalSecret) {
-    const { uid, name, namespace } = externalSecret.metadata
+    const { uid, name, namespace, annotations = {} } = externalSecret.metadata
     // NOTE(jdaeli): hash this in case resource version becomes too long?
     const secretDescriptor = { ...externalSecret.secretDescriptor, name }
     const ownerReference = {
@@ -47,7 +47,7 @@ class Daemon {
       uid
     }
 
-    return { id: uid, namespace, secretDescriptor, ownerReference }
+    return { id: uid, namespace, secretDescriptor, ownerReference, annotations }
   }
 
   /**
@@ -75,6 +75,7 @@ class Daemon {
       kubeClient: this._kubeClient,
       logger: this._logger,
       namespace: descriptor.namespace,
+      annotations: descriptor.annotations,
       secretDescriptor: descriptor.secretDescriptor,
       ownerReference: descriptor.ownerReference
     })
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -24,6 +24,7 @@ class Poller {
    * @param {SecretDescriptor} secretDescriptor - Kubernetes secret descriptor.
    */
   constructor ({
+    annotations = {},
     backends,
     intervalMilliseconds,
     kubeClient,
@@ -32,6 +33,7 @@ class Poller {
     secretDescriptor,
     ownerReference
   }) {
+    this._annotations = annotations
     this._backends = backends
     this._intervalMilliseconds = intervalMilliseconds
     this._kubeClient = kubeClient
@@ -55,6 +57,7 @@ class Poller {
       apiVersion: 'v1',
       kind: 'Secret',
       metadata: {
+        annotations: this._annotations,
         name: secretDescriptor.name,
         ownerReferences: [
           this._ownerReference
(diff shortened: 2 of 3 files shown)


jeffpearce:
Here's the failing test (from the Travis build).

image

I think that the annotations should be optional; if not provided in the external secret, we shouldn't try to pass them when we create the secret. That would probably fix that failing test.

Also, rather than modifying the existing tests, I think we should just add a new one that verifies that the code can pass annotations (if provided). The unmodified existing tests would verify that the behavior hasn't changed if there are no annotations.

Let me know if you have questions about any of that.

cabrinha:
> Thanks for the PR @cabrinha
> 
> Here's the failing test (from the Travis build).
> 
> image
> 
> I think that the annotations should be optional; if not provided in the external secret, we shouldn't try to pass them when we create the secret. That would probably fix that failing test.
> 
> Also, rather than modifying the existing tests, I think we should just add a new one that verifies that the code can pass annotations (if provided). The unmodified existing tests would verify that the behavior hasn't changed if there are no annotations.
> 
> Let me know if you have questions about any of that.

Yep, thanks for taking a look @jeffpearce ! I changed annotations to be optional and also altered the existing tests to expect an empty annotations key. If I need to add specific tests let me know, otherwise I think this could be good to go now.

lib/daemon.js
```diff
@@ -36,7 +36,7 @@ class Daemon {
    * @returns {Object} Poller descriptor.
    */
   _createPollerDescriptor (externalSecret) {
-    const { uid, name, namespace } = externalSecret.metadata
+    const { uid, name, namespace, annotations = {} } = externalSecret.metadata
```
silasbw: In general, I worry about copying one object's annotations directly to another object.  A more "kubernetes-like" approach would be to have a `Secret` template (analogous to the `Pod` template for `Deployment`s) and set the annotation values there that you want the external secrets controller to apply to the corresponding `Secret`.

We've discussed this approach before, but had some concerns with reconciling in with other "frontends" that don't use `Secret` objects (see https://github.com/godaddy/kubernetes-external-secrets/issues/46).

What's the use case for this?  Maybe there's another way to support it?

cabrinha:
@silasbw thanks for the consideration. The use-case is covered in https://github.com/godaddy/kubernetes-external-secrets/issues/137.

I'd like to use the [kubernetes-replicator](https://github.com/mittwald/kubernetes-replicator) deployment to copy external secrets (and maybe later configmaps) to other/all namespaces automatically.

What would be better is if I didn't have to use that project at all, but if there were an option to have the ExternalSecret replicated to every (or some) namespace on creation.

Create the secret once, but be able to consume it everywhere.

silasbw:
This fell off my radar @cabrinha.  Is it still useful for you?  If it us, would something like the approach I mention above work?  For example, there'd be a `secretDescriptor.template.metadata.annotations` object you could add annotations to that the controller would copy to your resulting `Secret` object.

cabrinha:
> This fell off my radar @cabrinha. Is it still useful for you? If it us, would something like the approach I mention above work? For example, there'd be a `secretDescriptor.template.metadata.annotations` object you could add annotations to that the controller would copy to your resulting `Secret` object.

That would be great. Yes, this feature would still be useful to me. I think others would also find a use for it along side [kubernetes-replicator](https://github.com/mittwald/kubernetes-replicator).

Let me know if there is anything I can do to help.

iAnomaly:
This is critical for us as well as we use Helm hook annotations on our Secrets currently. @silasbw @cabrinha What needs to happen to get this merged and how can I help? Thanks.

silasbw:
I think `secretDescriptor.template.metadata.annotations` is the way to go.  An `ExternalSecret` using that might look like this, for example:

```yaml
kind: ExternalSecret
metadata:
  name: hello-service
secretDescriptor:
  backendType: secretsManager
  data:
    - key: hello-service/password
      name: password
  template:
    metadata:
      labels:
        dog: farfel
      annotations:
        cat: cheese
```

The corresponding `Secret` would look like:

```yaml
kind: Secret
metadata:
  labels:
    dog: farfel
  annotations:
    cat: cheese
...
```

@iAnomaly, want to take a pass at implementing that?

iAnomaly:
@silasbw @cabrinha #192

silasbw:
Closing in favor of https://github.com/godaddy/kubernetes-external-secrets/pull/192

PR #156, opened 2019-08-30
Added parameter to control secrets backend polling

zdmytriv:
Trying to address AWS SSM rate limits issue. We are not planning to change/rotate secrets value on the fly. Will are planning to address secret rotation by updating versioning into secret name.

Better fix for https://github.com/godaddy/kubernetes-external-secrets/issues/117

co-author @max-lobur, please review

### DIFF
--- a/charts/kubernetes-external-secrets/templates/deployment.yaml
+++ b/charts/kubernetes-external-secrets/templates/deployment.yaml
@@ -32,10 +32,8 @@ spec:
             {{- toYaml .Values.resources | nindent 12 }}
           env:
           {{- range $name, $value := .Values.env }}
-          {{- if not (empty $value) }}
           - name: {{ $name | quote }}
             value: {{ $value | quote }}
-          {{- end }}
           {{- end }}
           # Params for env vars populated from k8s secrets
           {{- range $key, $value := .Values.envVarsFromSecret }}
--- a/charts/kubernetes-external-secrets/templates/rbac.yaml
+++ b/charts/kubernetes-external-secrets/templates/rbac.yaml
@@ -11,7 +11,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["secrets"]
-    verbs: ["create", "update"]
+    verbs: ["get", "create", "update"]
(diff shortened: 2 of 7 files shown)


charts/kubernetes-external-secrets/templates/deployment.yaml
```diff
@@ -32,11 +32,9 @@ spec:
             {{- toYaml .Values.resources | nindent 12 }}
           env:
           {{- range $name, $value := .Values.env }}
-          {{- if not (empty $value) }}
```
zdmytriv: with this env variable like this `MY_VAR=false` wouldn't be taking into account

charts/kubernetes-external-secrets/templates/rbac.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["secrets"]
-    verbs: ["create", "update"]
+    verbs: ["get", "create", "update"]
```
zdmytriv: I was getting `403` here https://github.com/godaddy/kubernetes-external-secrets/blob/master/lib/poller.js#L105

Flydiverny:
Not sure I'd call this a better fix for #117 as it completely disables polling and updating on changes. Seems a bit counterintuitive to _never_ have the created secret updated once its been created from the external secret. Even if the external secret is modified you would need to manually remove the secret that was created before hand to make sure it's recreated by the modification.

I'm all for an option to disable the polling, seems like that could make sense for some, but I'd at least want it to update my secret if i modify the external secret.

silasbw:
@Flydiverny how should we reconcile this with your recent work?  Do you think it adds value to https://github.com/godaddy/kubernetes-external-secrets/pull/133 ?

Flydiverny:
> @Flydiverny how should we reconcile this with your recent work? Do you think it adds value to #133 ?

I guess an alternative to disable polling intervall could still make sense, so we could look at generation changes to know when to update the secret, ie edited external secret means updated secrets. So no automatic updates if one were to change the value in the backend. 

As a semi work around I guess you should now be able to set a polling interval to a very long time :smile:

PR #163, opened 2019-09-17
Added annotation section in the deployment descriptor to set the isti…

saqibsaleem:
…o sidecar injection.

Added value in the values file to set the value to inject.


Flydiverny:
Why not just set 

```yaml
podAnnotations:
  sidecar.istio.io/inject: "true"
```

in your values when needed?

iAnomaly:
Agreed with @Flydiverny, the functionality to accomplish this already exists and should be used. @silasbw, can this be closed?

Flydiverny:
Closing this, if there's any issue or reason why the podAnnotations used as above wouldn't be sufficient or work please open an issue! :)

PR #171, opened 2019-10-11
Add external-secrets.yml patching to release.sh

reegnz:
Also bumped the version to the latest released version manually.
Also changed Deployment apiVersion because extensions/v1beta1 is
deprecated.


reegnz:
Adds the patching of the yaml on release mentioned in #89.

Issue #160, opened 2019-09-09
serviceAccount.name not used by deployment in the helm chart


zarrarrana:
Hi, 
serviceAccount.name  is mentioned as a parameter, but not used in the deployment.
https://github.com/godaddy/kubernetes-external-secrets/blob/master/charts/kubernetes-external-secrets/templates/deployment.yaml#L26

Zarrar

Issue #161, opened 2019-09-09
Upgrade aws-sdk to support EKS IAM Roles for Service Accounts


zarrarrana:
Update aws-sdk to 2.521.0 or higher to support  EKS IAM Roles for Service Accounts.

Zarrar

marcincuber:
Definitely waiting for this upgrade!

derrickburns:
Unfortunately, just upgrading the aws-sdk does not appear to be sufficient.  

I built the package with 
```
 "aws-sdk": {
      "version": "2.528.0",
      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.528.0.tgz",
      "integrity": "sha512-WBgavPqKHvYcIhD7LhmLLYOKtG/SUdNY6hMB0N/Jxaolzx4sOa7xegiZ8sdkAUq/vUEs1frnBF9hNmeIszUWCg==",
```

And deployed it.  I get:

```
{"level":50,"time":1568607194719,"pid":16,"hostname":"external-secrets-59f8c4d964-wn59s","message":"User: arn:aws:sts::118346523422:assumed-role/eksctl-test1- │
│ nodegroup-ng-NodeInstanceRole-F9L42IA4PBFA/i-07f29f17597c840c1 is not authorized to perform: ...
```

Note the role is WRONG.  It should be the role from the environment:

```
│    Environment:                                                                                                                                                ││      AWS_REGION:                    us-west-2                                                                                                                  ││      POLLER_INTERVAL_MILLISECONDS:  10000                                                                                                                      ││      AWS_ROLE_ARN:                  arn:aws:iam::118346523422:role/eksctl-test1-addon-iamserviceaccount-externa-Role1-NZBWIFHK6XY4
```

marcincuber:
@derrickburns I did exactly the same and basically temporary credentials are not provisioned for that provided service accounts. Do you think it might be an issue with aws-sdk for Node?

derrickburns:
@marcincuber That is my guess.

derrickburns:
So we have a workaround but we cannot implement it with this helm chart.  Can you allow one to set the securityContext in the deployment?

marcincuber:
@derrickburns are you staying that adding the following solves the issue?
```
      securityContext:
        fsGroup: 65534
```

farshad-hobsons:
> @derrickburns are you saying that adding the following solves the issue?
> 
> ```
>       securityContext:
[...]

PR #172, opened 2019-10-14
Add support for serviceaccount annotations

### LINKED ISSUE #170: Add support to link existing service account in deployment.yaml
When deploying via helm using the following, the passed existing service-account(`um-sa-test`) is not used in deployment.yaml. Rather it tries to use a non-existing service-acount called `um-kubernetes-external-secret-kubernetes-external-secrets`
```bash
helm  upgrade um-kubernetes-external-secret external-secrets/kubernetes-external-secrets --set serviceAccount.name=um-sa-test --set env.AWS_REGION=us-east-1 --set serviceAccount.create=false --version 1.0.1 --debug -i
```

The error generated as a result of a non-existing service accounts is:
```
29s         Warning   FailedCreate        replicaset/um-kubernetes-external-secret-kubernetes-external-secrets-59c68b976f   Error creating: pods "um-kubernetes-external-secret-kubernetes-external-secrets-59c68b976f-" is forbidden: error looking up service account default/um-kubernetes-external-secret-kubernetes-external-secrets: serviceaccount "um-kubernetes-external-secret-kubernetes-external-secrets" not found
```
https://github.com/godaddy/kubernetes-external-secrets/pull/169 fixes this issue of deployment with these paramaters `serviceAccount.create=false serviceAccount.name=um-sa-test `

sklemmer:
EKS supports specifying an IAM Role to a service account. (see #161 and https://docs.aws.amazon.com/en_pv/eks/latest/userguide/iam-roles-for-service-accounts.html)
Currently we have to create a seperate service account and link it to the deployment.

This PR adds support to add this annotation directly using the chart.

Should fix #170

### DIFF
--- a/charts/kubernetes-external-secrets/templates/deployment.yaml
+++ b/charts/kubernetes-external-secrets/templates/deployment.yaml
@@ -23,7 +23,7 @@ spec:
         {{- toYaml .Values.podAnnotations | nindent 8 }}
       {{- end }}
     spec:
-      serviceAccountName: {{ template "kubernetes-external-secrets.fullname" . }}
+      serviceAccountName: {{ template "kubernetes-external-secrets.serviceAccountName" . }}
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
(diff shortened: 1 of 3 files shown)


zarrarrana:
This should fix #160 also.

PR #179, opened 2019-10-19
fix: iam role

### LINKED ISSUE #174: Use of "iam.amazonaws.com/permitted" annotation causes conflicts with existing KIAM setups
## I'm submitting a...
* [X] bug report 

- related to `roleArn` functionality described in issue https://github.com/godaddy/kubernetes-external-secrets/issues/143
and implemented in PR https://github.com/godaddy/kubernetes-external-secrets/pull/144

- Use of KIAM `iam.amazonaws.com/permitted` annotation for `roleArn` validation causes conflicts with existing KIAM setups.

- `iam.amazonaws.com/permitted` is a KIAM created and managed annotation, and not part of any known standard.
(e.g. kube2iam uses a different annotation: `iam.amazonaws.com/allowed-roles` )

### What is the current behavior?

Kubernetes external secrets errors out when `iam.amazonaws.com/permitted` is present, has valid values intended for `kiam`, but is not provided a `roleArn` in the `ExternalSecret` descriptor.

This is a common use case when `kiam` is handling the IAM for the `kubernetes-external-secret` operator, and no additional role assumption is required by the individual `ExternalSecret` resources

To reproduce:

``` 
kind: ExternalSecret
metadata:
  annotations:
  name: test-service
secretDescriptor:
 #No Role ARN Provided
  backendType: secretsManager
  data:
  - key: test-service/password
    name: password
```

Namespace annotated for `kiam` support only:

```
kind: Namespace
metadata:
  annotations:
    iam.amazonaws.com/permitted: my-kiam-roles.* 
```

``` 
{"level":50,"time":XXX,"pid":16,"hostname":"XXX","type":"Error","stack":"Error: not allowed to fetch secret: test-service: namspace does not allow to assume role undefined
    at Poller._upsertKubernetesSecret (/app/lib/poller.js:110:14)
    at process._tickCallback (internal/process/next_tick.js:68:7)","msg":"failure while polling the secret test-service","v":1}
```

### What's the expected behavior?

Secret sync should be allowed when namespace annotation is present, and no `roleArn` is provided.

```         
{
     // test undefined
     ns: { metadata: { annotations: { 'iam.amazonaws.com/permitted': 'my-kiam-role.*' } } },
     descriptor: { roleArn: undefined },
     permitted: true
}, 
```

### What is the current behavior?

`kiam` or `kubernetes-external-secrets` will now **incorrectly permit** (i.e. pass the regex and proceed to attempt an sts call) assumption of roles in a namespace not intended to be used by `kiam` or `kubernetes-external-secrets`.  The sharing of an annotation allows for nondeterministic behavior from each operator when used at the same time.  

Scenario (AWS ARNs truncated for brevity):

- `kubernetes external secret` controller uses `kiam` to assume roles for processing, with that role (and all other allowed `kiam` roles) covered by the `iam.amazonaws.com/permitted: kiam-roles.*` annotation on the namespace

- `kubernetes external secrets` `ExternalSecret` has a desired `roleArn` value `my-team-external-secret-role.*`

- `kiam` policies are attached to `kiam-server` pods, and have no knowledge of `ExternalSecret` level `kubernetes-external-secret` role / polices 

- `kubernetes-external-secret` policies are assumed at runtime from the `kubernetes-external-secret` controller AFTER it assumes a role via `kiam`, and have no knowledge of all the other `kiam` level role / polices 

In order to support this, one of the following must now happen:

- regex updated to support both patterns: `(kiam|my-team-external-secret-role).*`
  - now `kiam` OR `kubernetes-external-secrets` is _"permitting"_ roles they can not actually assume 
- naming convention updated to support a single shared pattern: `namespace-iam-roles.*` 
  - same result, `kiam` OR `kubernetes-external-secrets` is still _"permitting"_ roles they can not actually assume 

### What's the expected behavior?

One of the following should be done to `kubernetes-external-secrets` prevent conflicts
- Backwards Compatible Break - change the support annotation for  `kubernetes-external-secrets` to be something under the existing CRD naming of `externalsecrets.kubernetes-client.io`, 

- Allow the annotation to be configurable so that both `kiam` and `kubernetes-external-secrets` can run without conflict

- Explicitly document this issue in the `README.md`

moolen:
This PR fixes #174:

* (bug) sync should work when namespace annotation is present, and no roleArn is provided
* (feature) make the role-permission annotation key configurable

These issues are addressed in two separate commits.

### DIFF
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -12,8 +12,6 @@
  *   object, this is the property name of the value to use.
  */
 
-const annotationPermittedKey = 'iam.amazonaws.com/permitted'
-
 /** Poller class. */
 class Poller {
   /**
@@ -25,6 +23,7 @@ class Poller {
    * @param {string} namespace - Kubernetes namespace.
    * @param {Object} customResourceManifest - CRD manifest
    * @param {Object} externalSecret - ExternalSecret manifest.
+   * @param {string} rolePermittedAnnotation - namespace annotation that defines which roles can be assumed within this namespace
    * @param {Object} metrics - Metrics client.
    */
   constructor ({
@@ -34,6 +33,7 @@ class Poller {
     logger,
     metrics,
     customResourceManifest,
+    rolePermittedAnnotation,
     externalSecret
   }) {
     this._backends = backends
@@ -42,6 +42,7 @@ class Poller {
     this._logger = logger
     this._timeoutId = null
     this._metrics = metrics
+    this._rolePermittedAnnotation = rolePermittedAnnotation
     this._customResourceManifest = customResourceManifest
 
     this._externalSecret = externalSecret
@@ -175,7 +176,7 @@ class Poller {
       }
     }
     // an empty annotation value allows access to all roles
-    const re = new RegExp(namespace.metadata.annotations[annotationPermittedKey])
+    const re = new RegExp(namespace.metadata.annotations[this._rolePermittedAnnotation])
 
     if (!re.test(role)) {
       allowed = false
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -169,7 +169,7 @@ class Poller {
     let allowed = true
     let reason = ''
 
-    if (!namespace.metadata.annotations) {
+    if (!namespace.metadata.annotations || !role) {
       return {
         allowed, reason
       }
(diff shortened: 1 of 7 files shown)


lib/poller.js
```diff
@@ -130,13 +131,19 @@ class Poller {
     let allowed = true
     let reason = ''
 
+    if (!role) {
+      return {
+        allowed, reason
+      }
+    }
```
Flydiverny: might as well merge thisinto the if below

moolen: yes, that makes sense, thanks!

Issue #175, opened 2019-10-14
Versioning support in Secrets Manager


furikake:
AWS Secrets Manager supports versioning and you can retrieve the appropriate version via the `VersionStage` or `VersionId` properties. https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SecretsManager.html#getSecretValue-property.

The default behaviour is to retrieve the version with `VersionStage: 'AWSCURRENT'`.  I would like to retrieve the previous version `AWSPREVIOUS` of the secret also in order to support zero downtime rotation of secrets.  I don't have a need to support `VersionId` but others might.

I can't think of a neat way to support describing versioning given the various backends and even Secrets Manager has two methods.  An example of how this could be described:

```
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  name: somesaas-external-secret
secretDescriptor:
  backendType: secretsManager
  data:
    - key: prod/somesaas
      name: somesaas_client_id_previous
      property: client_id
      version: AWSPREVIOUS
    - key: prod/somesaas
      name: somesaas_client_id_current
      property: client_id
```

Is this something that you are working on?  I can have a go at creating a PR for it if you don't have plans and don't mind supporting this feature.

Flydiverny:
#181 was merged with `versionStage` option for keys and will be available in next release.

PR #181, opened 2019-10-22
Added support for secrets versioning in Secrets Manager using version stage labels

furikake:
This commit is to add versioning support as per #175


furikake:
I figured most people would be labeling their versions with a staging label instead of relying on a version id (UUID) since it's difficult to work with.  Happy to update though.

Do you think it `version` should be renamed:

```
      name: password
      # Version Stage in Secrets Manager
      versionStage: AWSPREVIOUS
      property: password_previous
```

or nested under `version.versionStage`

```
      name: password
      # Version Stage in Secrets Manager
      version: 
        versionStage: AWSPREVIOUS
      property: password_previous
```

Flydiverny:
@furikake I think just renaming it to `versionStage` is fine :)

PR #184, opened 2019-10-25
numbers secret properties to string

juancgalvis:
parse number value to string to fix error on create secret with number value on k8s

### DIFF
--- a/lib/backends/kv-backend.js
+++ b/lib/backends/kv-backend.js
@@ -43,7 +43,11 @@ class KVBackend extends AbstractBackend {
           throw new Error(`Could not find property ${secretProperty.property} in ${secretProperty.key}`)
         }
 
-        return parsedValue[secretProperty.property]
+        let propertyValue = parsedValue[secretProperty.property]
+        if (typeof propertyValue === 'number') {
+          propertyValue = `${propertyValue}`
+        }
+        return propertyValue
       }
 
       return value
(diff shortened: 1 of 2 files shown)


lib/backends/kv-backend.js
```diff
@@ -43,7 +43,11 @@ class KVBackend extends AbstractBackend {
           throw new Error(`Could not find property ${secretProperty.property} in ${secretProperty.key}`)
         }
 
-        return parsedValue[secretProperty.property]
+        let propertyValue = parsedValue[secretProperty.property]
+        if (typeof propertyValue === 'number') {
+          propertyValue = `${propertyValue}`
+        }
+        return propertyValue
```
Flydiverny: So the issue here is that we can't base64 encode it yes?
Perhaps we can change
https://github.com/godaddy/kubernetes-external-secrets/blob/088efdfa2c21f8e2c81313e0d4636409aabe53e6/lib/backends/kv-backend.js#L78

To
```js
 data[secretProperty.name] = (Buffer.from(`${secretPropertyValues[index]}`, 'utf8')).toString('base64') 
```

So its possible to set boolean and numbers and they'll all still be stringified so they can get encoded for the secret. This will support `null` as well, not sure if thats an issue but kind of seems like an edge case. We could check if propertyValue is null and throw an error

juancgalvis: Yes, this is the issue, so I will update the pull request with your suggestions.

Flydiverny: @juancgalvis I included this in #196 as I was refactoring when adding support for `dataFrom`  which is now merged and will be available in next release! :)

Issue #94, opened 2019-06-18
feature request: create secret with every entry in secret manager key


shraykay:
Would it be possible to add an extra section under External Secrets that will just import all key/values under the stored AWS key and add them to a Kubernetes secret?

I would like to avoid having to explicitly list out every key and property every time I update or add a new k/v to a AWS secret.

silasbw:
We've been waffling on adding options for more "advanced" processing like this.  Mostly because we don't understand the impact/utility of adding them (and we're worried about the longer term maintenance cost), and how general to make the advanced options so that they might scale to backends we add in the future.

Can you help us understand the impact for you?  How many key, property objects do you have in an ExternalSecret?  How many ExternalSecrets are you juggling?

shraykay:
we have about 50 ExternalSecrets with ~15 key value pairs within each. We primarily use the envFrom feature in deployments to be able to source directly without having to specify each one by one.

silasbw:
OK, thank you.  It seems like this is worth considering supporting.  We should think through _how_.  *e.g.*, single entry in the `secretDescriptor.data` array?  some property on `secretDescriptor` that's mutually exclusive with `secretDescriptor.data`, or do we support some well-defined merge behavior?  ...  We'd welcome some concrete suggestions / brainstorming :)

shraykay:
I think introducing another field would be great, similar to envFrom (especially for maintaining backwards). for "well defined merge behavior", anything that's manually specified in the data area should override anything that's in the envFrom area. 

Thank you for responding so quickly!

Flydiverny:
I think having a basic edition `dataFrom` would cover most common use cases? To make it slightly more advanced you could allow merge / overrides from data entries.

```yaml
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  name: postgres-credentials
secretDescriptor:
  backendType: secretsManager
  dataFrom: /development/postgres/credentials
```

secrets manager secret
```json
{
    "Name": "/development/postgres/credentials",
    "SecretString":  "{\"username\":\"my-pg-user\", \"password\":\"my-super-secret-password\"}",
}
```

Resulting secret
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: postgres-credentials
type: Opaque
data:
  username: bXktcGctdXNlcg== # my-pg-user
  password: bXktc3VwZXItc2VjcmV0LXBhc3N3b3Jk # my-super-secret-password
```

----

[...]

PR #189, opened 2019-10-30
Add feature for fetching all JSON secret properties.

iAnomaly:


arruzk:
@iAnomaly this change has a bug when you specify `property` option.
```
aws secretsmanager create-secret --name hello-service/credentials --secret-string '{"username":"admin","password":"1234"}'
```
```
kind: ExternalSecret
metadata:
  name: test-secrets-notifier
secretDescriptor:
  backendType: secretsManager
  data:
      name: password
      property: password
```
Result is (instead of `1234`):
```
{
    "apiVersion": "v1",
    "data": {
        "0": "MQ==",
        "1": "Mg==",
        "2": "Mw==",
        "3": "NA=="
    },
    "kind": "Secret",
...
```

iAnomaly:
@arruzk Thanks for finding and reporting but I cannot reproduce no matter how hard I try. What version of node did you test this on? Only thing I can think of is Objects.keys changed for newer versions of ECMA so I switched to `Object.getOwnPropertyNames` if you don't mind testing again.

Also expanded the test coverage and fixed one bug corner case (not the one you reported).

arruzk:
@iAnomaly thank you! 

Now it's working fine if I specify both `name` and `property`. 
I have an error if I specify only `property` (without `name`), but probably it should be like this (maybe error can be a little more verbose)
```
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be one of type string, Buffer, ArrayBuffer, Array, or Array-like Object. Received type object\n    at Function.from (buffer.js:233:9)\n    at Object.getOwnPropertyNames.forEach.key (/Users/u/projects/tmp/kubernetes-external-secrets/lib/backends/kv-backend.js:82:29)\n    at Array.forEach (<anonymous>)\n    at externalData.forEach (/Users/u/projects/tmp/kubernetes-external-secrets/lib/backends/kv-backend.js:81:63)\n    at Array.forEach (<anonymous>)\n    at SecretsManagerBackend.getSecretManifestData (/Users/u/projects/tmp/kubernetes-external-secrets/lib/backends/kv-backend.js:75:18)\n    at process._tickCallback (internal/process/next_tick.js:68:7)","msg":"failure while polling the secret test-secrets-notifier","v":1}
```
```
  data:
    - key: staging/secrets-common
      property: SECRET_HASH_SALT
```
`secretPropertyValues` object is 
```
[{ undefined: { SECRET_HASH_SALT: 'salt' } }]
```
Node version: v10.13.0
Npm version: 6.4.1

iAnomaly:
Thank you @arruzk, I could reproduce this one. It is now fixed with added test coverage for this case. Appreciate your testing!

Flydiverny:
To add this I think we should do something similar to what was discussed here https://github.com/godaddy/kubernetes-external-secrets/issues/94#issuecomment-505290048

Flydiverny:
I took the liberty of taking a pass on this in #195 ( based on #94 )

iAnomaly:
Closed in favor of #196.

PR #192, opened 2019-11-04
Secret annotations passthrough

iAnomaly:


Flydiverny:
We should probably add some docs for this :)

lib/poller.test.js
```diff
@@ -149,6 +149,91 @@ describe('Poller', () => {
...
+          annotations: [
+            {
+              cat: 'cheese'
+            }
+          ],
+          labels: [
+            {
+              dog: 'farfel'
+            }
+          ],
```
Flydiverny: Should this really be arrays? I'd expect this to be objects? :)

```js
metadata: {
  annotations: {
    cat: 'cheese'
  },
  labels: {
    dog: 'farfel'
  },
}
```

lib/poller.js
```diff
@@ -67,6 +70,12 @@ class Poller {
       type: secretDescriptor.type || 'Opaque',
       data
     }
+
+    if (template) {
+      secretManifest = merge(clonedeep(secretManifest), template, secretManifest)
```
Flydiverny: if we give template a default value you could skip the if check and return the merge result instead. 

Also should be fine to remove one of the merges:
```js
return merge(clonedeep(template), secretManifest)
```

iAnomaly: I prefer this way. I don't want to deep clone or merge unless we need to even if template had an empty object default value. 

The seemingly duplicate/two merges are there intentionally to avoid overwriting existing field values. I cover this in the docs.

iAnomaly:
@Flydiverny @silasbw Docs added and feedback resolved.

lib/poller.js
Flydiverny: I agree that doing the following would be a bad choice: 
```js
merge(clonedeep(secretManifest), template)
```

But I have a hard time seeing the benefit of this
```js
merge(clonedeep(secretManifest), template, secretManifest)
```
over this
```js
merge(clonedeep(template), secretManifest)
```

Test snippet:
```js
const _ = require('lodash');

const objA = { myObj: "some", annotations: {} }
const template = { annotations: [] }

// A <- template <- A
const altA = _.merge(_.cloneDeep(objA), template, objA);

// template <- A
const altB = _.merge(_.cloneDeep(template), objA);

console.log('Are equal:', _.isEqual(altA, altB), altA, altB)
```
https://repl.it/repls/WrathfulMessyMicrostation

Both alternatives seem equally bad at handling invalid templates, maybe I'm missing something? :) Do you have an example for when one would work but not the other?

iAnomaly: Ah, I read too quickly and misunderstood. I thought you were advocating for 
```javascript
merge(clonedeep(secretManifest), template)
```
Agreed this one is better:
```javascript
merge(clonedeep(template), secretManifest)
```
Will fix.

iAnomaly: Ah, I read too quickly and misunderstood. I thought you were advocating for 
```javascript
merge(clonedeep(secretManifest), template)
```
Agreed this one is better:
```javascript
merge(clonedeep(template), secretManifest)
```
Will fix.

PR #193, opened 2019-11-04
feat: allow limiting scope to one namespace

### LINKED ISSUE #106: RFE: Restrict watching to a single namespace.

Flydiverny:
# If you have the time feel free to make a new PR for this :) JeroenRijks had some additional suggestions here   https://github.com/Flydiverny/kubernetes-external-secrets/pull/1/ Do note that this base is out of date 

Allows watching of a single namespace by setting env `SCOPE_NAMESPACE`  instead of watching all namespaces.

Fixes #106 

Things to fix/consider:

- [ ] tests
- [ ] docs
- [ ] If I am running multiple external secrets controllers, should they all be creating the CRD?
- [ ] Limited RBAC if namespaced, no need to get access to everything.


JeroenRijks:
Hi, I'd love to see this feature pushed into master, but I've not got much experience on the javascript side of things, I'm on the k8s side. Is there any way I can be of use to get this through, e.g. manual testing?

silasbw:
@Flydiverny is this ready for a review?

Flydiverny:
@silasbw I mean it's functional in itself, so I suppose rest could be improved afterwards

Flydiverny:
Wont have time to wrap this up for awhile so if we want to merge it as is thats fine by me, or if we want to wait for tests etc, someone is welcome to take it over! Or I'll hopefully get to it soon enough :D

lib/external-secret.js
```diff
@@ -26,18 +26,23 @@ async function startWatcher ({
...
+  scopeNamespace = null
 }) {
   const deathQueue = createEventQueue()
 
+  let watchEndpoint = kubeClient.apis[customResourceManifest.spec.group]
+    .v1.watch[customResourceManifest.spec.names.plural]
+
+  if (scopeNamespace) {
+    watchEndpoint = kubeClient.apis[customResourceManifest.spec.group]
+      .v1.watch.namespaces(scopeNamespace)[customResourceManifest.spec.names.plural]
```
JeroenRijks: :+1:

JeroenRijks:
@silasbw - any chance you could review/merge this? I've not got write permissions

ainformatico:
We are having the same issue, is there any progress on getting this PR merged?

Flydiverny:
If anyone has the extra time, feel free to create an updated PR which addresses the checkboxes in the description :) This is still a few steps down my list.

JeroenRijks:
I've addressed RBAC and documentation, but not tests or the multiple CRDs approach. 
https://github.com/Flydiverny/kubernetes-external-secrets/pull/1

About CRDs - do we think only one CRD should be defined across the cluster, or should they be namespaced? 
My problems with limiting CRD definitions to a namespace are:
 -  I don't see a situation where users will want one version of the CRD in one namespace, and another version in another namespace.
 - The CRD is not currently included as part of the chart. If we were to limit its scope to a namespace, we'd have to add it to the chart, meaning that this (relatively small) change will affect users that don't even use the feature

Flydiverny:
@JeroenRijks thanks for taking the time!

Regarding CRDs they will be cluster scoped, so someone running multiple versions of external-secrets in their cluster will have to make sure that the CRD versions are compatible. 

I think we will have to add an option to enable / disable creation of the CRD tho to give better control since currently the CRD is always upserted on start of external-secrets.

JeroenRijks:
No problem @Flydiverny , thanks for making the PR to begin with. 

I've pushed a new commit to my PR (from my fork into your fork). If the DO_NOT_UPDATE_CRD env variable is set, the pod will check if the CRD already exists, and will not update it if it exists. If it doesn't exist, it will create it. This variable is not set by default, and requires users to set `scopeNamespace.doNotUpdateCrd=true` when installing the chart. Please make changes, or let me know what changes to make, if there's anything wrong - I'm not great with JS

JeroenRijks:
I can't figure out the right way to update/create the CRD when Roles are used instead of ClusterRoles. If you set `scopeNamespace.enabled=true` and `scopeNamespace.useExistingCrd=true` in [my PR](https://github.com/Flydiverny/kubernetes-external-secrets/pull/1/), the release uses a Role. However, it can't [verify that the CRD exists](https://github.com/Flydiverny/kubernetes-external-secrets/pull/1/files#diff-c5ccca10391f649a523e175e555dc509R118) unless the release uses a ClusterRole.

I can't get further with this - could someone else try checking if the CRD exists, without granting cluster-wide access to the release? If not, I suggest that we always use ClusterRoles, regardless of whether the release is namespaced.

After this is done, and some tests are written, this should be ready to merge.

lib/external-secret.js
```diff
@@ -26,18 +26,23 @@ async function startWatcher ({
   kubeClient,
   customResourceManifest,
   logger,
-  eventQueue
+  eventQueue,
+  scopeNamespace = null
```
JacopoDaeli: Do we need the `null` default here?

Flydiverny: We don't necessarily need it but I think its cleaner to show that its optional and might be null instead of having it end up being `undefined`

Flydiverny:
> Correct me if I am wrong: if `SCOPE_NAMESPACE` env var is not assigned, then everything keeps working like it's currently working; if you assign a value to `SCOPE_NAMESPACE`, then we're limiting scope to the namespace specified via `SCOPE_NAMESPACE`.
> 
> Is it correct?

Yes, that's the intention.

> Also, something to clarify: what happen if the value assigned `SCOPE_NAMESPACE` is not a valid namespace?

This is a fair question, have not evaluated this scenario :) I believe it would crash due to the watch stream not finding the namespace in question, that or nothing at all will happen. Which behaviour would we like here? An intentional crash with a helpful error message?

sarmad-abualkaz:
What is still required for this PR to be merged? Any more testing or help needed on this @Flydiverny?

etienneszott0000:
We are having the same issue, is there any progress on getting this PR merged?

Flydiverny:
Regrettable this PR ended up on the backburner last time I had some time to put into KES.

I'm afraid I have little time to put here at the moment! :( So I can't make any promises on when I can catch up on this issue.

kuuji:
Hey @Flydiverny. Is there anything we can do to take this off your hands and help move it forward? 
What is left to do on this?
Is it only the handling of invalid `SCOPE_NAMESPACE`? And maybe some tests?

omermishania:
Are there any updates? Does this or something similar will be integrated?
Thanks in advance ;)

Flydiverny:
#548 was merged

PR #176, opened 2019-10-15
feat: integration of AWS IAM Roles for ServiceAccounts (IRSA)

farshad-hobsons:
### Summary of Changes:

- BUG: Add functionality to use an existing service account - Same as https://github.com/godaddy/kubernetes-external-secrets/issues/170 &  https://github.com/godaddy/kubernetes-external-secrets/pull/169
TESTED with:
```
helm --tiller-namespace ops --namespace ops upgrade um-kubernetes-external-secret --set serviceAccount.name=um-farshad-test --set env.AWS_REGION=us-east-1 --set serviceAccount.create=false  --debug -i ./charts/kubernetes-external-secrets
```

BEFORE change incorrect serviceAccountName in rendered yaml:
```yaml
...
# Source: kubernetes-external-secrets/templates/deployment.yaml
...
    spec:
      serviceAccountName: um-kubernetes-external-secret-kubernetes-external-secrets
```
AFTER change correct serviceAccountName in rendered yaml:
```yaml
...
# Source: kubernetes-external-secrets/templates/deployment.yaml
...
    spec:
      serviceAccountName: um-farshad-test
```
Also TESTED with (for regression purposes):
```
helm --tiller-namespace ops --namespace ops upgrade um-kubernetes-external-secret  --set env.AWS_REGION=us-east-1  --debug -i ./charts/kubernetes-external-secrets
```

- BUG: Latest [1.5.0 version requires `GET` on namespace](https://github.com/godaddy/kubernetes-external-secrets/blob/58283fd6c6114ad3e19bf6f51da636ff63dfbc76/lib/poller.js#L107) which the current chart was not providing.

- FEATURE: Optionally Parameterize `securityContext`(needed for IRSA support [see reference issue](https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8)) and `saAnnotations`(needed for [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html)) and `imagePullSecrets`(needed if someone wants to build their own image and deploy from a private repo)
TESTED with:
```
    helm --tiller-namespace ops --namespace ops upgrade um-kubernetes-external-secret --set env.AWS_REGION=us-east-1 --set securityContext.fsGroup=65534  --debug -i ./charts/kubernetes-external-secrets
[...]

PR #200, opened 2019-11-07
Integration with AWS IAM Roles for ServiceAccounts

arruzk:
It is a second attempt to do it (first is https://github.com/godaddy/kubernetes-external-secrets/pull/176). With recent changes it is easier.
- AWS SDK version was updated
- securityContext parameter in a helm chart was added

It should fix this issue https://github.com/godaddy/kubernetes-external-secrets/issues/161 

Sorry @farshad-hobsons if you are working on it now.
@Flydiverny @silasbw could you please review it?

### DIFF
--- a/charts/kubernetes-external-secrets/README.md
+++ b/charts/kubernetes-external-secrets/README.md
@@ -22,6 +22,12 @@ $ helm install --name my-release external-secrets/kubernetes-external-secrets
 
 > **Tip:** A namespace can be specified by the `Helm` option '`--namespace kube-external-secrets`'
 
+To install the chart with [AWS IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html):
+
+```bash
+$ helm install --name my-release --set securityContext.fsGroup=65534 --set serviceAccount.annotations."eks\.amazonaws\.com/role-arn"='arn:aws:iam::111111111111:role/ROLENAME' external-secrets/kubernetes-external-secrets
+```
+
 ## Uninstalling the Chart
 
 To uninstall/delete the deployment:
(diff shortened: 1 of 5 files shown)


imaffe:
@arruzk  Sorry for the late response, but I downloaded the latest version with 2.1.0
image
And the environment variable of this pod is 
image
(Which is set by the EKS IRSA)
But the logs shows it is still using the EC2(nodes') iam role, and is not working . (Sorry I cannot post the logs cuz it might be sensitive)

Have you tested if it is really working with the IRSA ? Thanks so much

arruzk:
@imaffe just checked now. Everything is working for me.
my test values.yaml (helm chart version is kubernetes-external-secrets-2.2.0)
```yaml
env:
  AWS_REGION: eu-west-1
# created in terraform
serviceAccount:
  name: secrets-manager
  create: false
image:
  tag: latest
  pullPolicy: Always
securityContext:
  fsGroup: 65534
```

From pod env:
```yaml
    - name: AWS_ROLE_ARN
      value: arn:aws:iam::11111111111:role/staging-ir-1-sm-reader-role
    - name: AWS_WEB_IDENTITY_TOKEN_FILE
      value: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

Try to log in into the server (with exec) and do 
`cat /var/run/secrets/eks.amazonaws.com/serviceaccount/token`

dudicoco:
@arruzk I am also experiencing the same issue as @imaffe.
I was able to exec into the container and cat the token.

Any ideas?

dudicoco:
@arruzk when testing this did you verify the role which is being assumed does not have the node in its trusted relationships? Is it possible the container was actually using the node's role + kiam in your tests?

imaffe:
> @arruzk I am also experiencing the same issue as @imaffe.
> I was able to exec into the container and cat the token.
> 
> Any ideas?

I've fixed it, and I checked the trust relationship in the EKS cluster page, seems my problem is the serviceaccount name defined doesn't align with the trusted IAM role arn in the trusted relationships. The service account name is not the same as the app name, I think it's their helm chart didn't follow the conventions.

Flydiverny:
Service account name can be overriden by providing `serviceAccount.name` in helm values. ( https://github.com/godaddy/kubernetes-external-secrets/blob/master/charts/kubernetes-external-secrets/values.yaml#L39) and how its set can be seen here https://github.com/godaddy/kubernetes-external-secrets/blob/master/charts/kubernetes-external-secrets/templates/_helpers.tpl#L36-L41

Flydiverny:
If there's specific steps to get it all working it would be great if someone wants to do a PR to update the docs :)

dudicoco:
Thanks @imaffe, turns out my namespace wasn't aligned in the trusted relationships of the IAM role.
@Flydiverny doesn't seem like there are special steps to get it working except for the usual steps needed to configure pod IAM role OIDC token, was user error on our part.
securityContext with fsGroup: 65534 is needed though, I'm not sure if it's documented?

Thanks for the help guys!

mo-saeed:
Anybody tried it with EKS ? the pod is always trying to use the node_group role although I have all configured right and this `cat /var/run/secrets/eks.amazonaws.com/serviceaccount/token` exists in the pod

dudicoco:
@mo-saeed yes we are using it with EKS, take a look at my previous comments in this issue.

PR #207, opened 2019-11-08
Feat: e2e tests

moolen:
Hey!
I'd like to add e2e tests to the repository in order to test the full path including helm chart configuration, AWS/Localstack integration and the good path (create ES, read data from SSM, wait for secret to appear). This is a first iteration, we can add more tests later on.

I also added validation to the `custom-resource-manifest.json` (using the new `.spec` field).

Unfortunately, we can't test the STS/assume-role integration - local stack does not seem to support it.

### DIFF
--- a/config/aws-config.js
+++ b/config/aws-config.js
@@ -2,23 +2,40 @@
 
 /* eslint-disable no-process-env */
 const AWS = require('aws-sdk')
+const clonedeep = require('lodash.clonedeep')
+const merge = require('lodash.merge')
 
 const localstack = process.env.LOCALSTACK || 0
 
-const secretsManagerConfig = localstack ? { endpoint: 'http://localhost:4584', region: 'us-west-2' } : {}
-const systemManagerConfig = localstack ? { endpoint: 'http://localhost:4583', region: 'us-west-2' } : {}
-const stsConfig = localstack ? { endpoint: 'http://localhost:4592', region: 'us-west-2' } : {}
+let secretsManagerConfig = {}
+let systemManagerConfig = {}
+let stsConfig = {}
+
+if (localstack) {
+  secretsManagerConfig = {
+    endpoint: process.env.LOCALSTACK_SM_URL || 'http://localhost:4584',
+    region: process.env.AWS_REGION || 'us-west-2'
+  }
+  systemManagerConfig = {
+    endpoint: process.env.LOCALSTACK_SSM_URL || 'http://localhost:4583',
+    region: process.env.AWS_REGION || 'us-west-2'
+  }
+  stsConfig = {
+    endpoint: process.env.LOCALSTACK_STS_URL || 'http://localhost:4592',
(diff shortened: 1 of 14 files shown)


Flydiverny:
Doh just added some crd validation things as well, just hadn't gotten to creating the pr 😅 (#208 208)

moolen:
Thanks, i'll remove my crd changes then.

config/aws-config.js
```diff
@@ -5,20 +5,35 @@ const AWS = require('aws-sdk')
...
+    region: process.env.AWS_REGION || 'us-west-2'
+  }
+}
 
 module.exports = {
-  secretsManagerFactory: (opts) => {
+  secretsManagerFactory: (opts = {}) => {
     if (localstack) {
-      opts = secretsManagerConfig
+      Object.assign(opts, secretsManagerConfig)
```
silasbw: nit: `opts = Object.assign({}, opts, secretsManagerConfig)`

I worry copying into `opts` directly could lead to bugs later.  *E.g.*, if the caller reuses the `opts` object it passes.

e2e/run-e2e-suite.sh
```diff
@@ -0,0 +1,112 @@
...
+  kind delete cluster \
+    --loglevel=${KIND_LOG_LEVEL} \
+    --name ${KIND_CLUSTER_NAME}
+
+}
+trap cleanup EXIT
+
+kubectl apply -f ${DIR}/localstack.deployment.yaml
+
+helm template $DIR/../charts/kubernetes-external-secrets \
```
silasbw: worth passing setting --name?  e.g., `helm template --name e2e`.

then you can avoid relying on `release-name` as the default prefix for the deployment name

e2e/run-e2e-suite.sh
```diff
@@ -0,0 +1,112 @@
...
+
+until kubectl get secret | grep -q ^external-secrets-e2e-token; do \
+  echo -e "waiting for api token"; \
+  sleep 3; \
+done
+
+echo -e "${BGREEN}Starting external-secrets e2e tests...${NC}"
+
+kubectl rollout status deploy/localstack
+kubectl rollout status deploy/release-name-kubernetes-external-secrets
```
silasbw: this would change to `kubectl rollout status deploy/e2e-kubernetes-external-secrets`

moolen:
Thanks for your comments! I changed the deployment name for more clarity and squashed my changes into a single commit.

config/index.js
```diff
@@ -39,6 +39,7 @@ const systemManagerBackend = new SystemManagerBackend({
   logger
 })
 const backends = {
+  // when adding a new backend, make sure to change the CRD property too
```
silasbw: is this comment still valid, or is it an artifact from a previous revision?

Flydiverny: its valid :) he had CRD validation in this PR first, but since I had a pr for it as well it was removed from here, but it still applies!

PR #208, opened 2019-11-08
feat: add validation to CRD

Flydiverny:


crd.yaml
```diff
@@ -0,0 +1,76 @@
...
+    name: Last Sync
+    type: date
+  - JSONPath: .status.status
+    name: status
+    type: string
+  - JSONPath: .metadata.creationTimestamp
+    name: Age
+    type: date
+
+  validation:
```
silasbw: do you know which version of Kubernetes introduced validation for CRDs?  I think we should use validation, but we might need to explicitly call out the minimum version required.

Flydiverny: Seems to be available in 1.12 which is as far back as the official hosted kubernetes docs go now
https://v1-12.docs.kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation 
think its a fair requirement :)

Flydiverny: Heres the list I was looking for https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
Seems to have been enabled by default as beta in 1.9

silasbw: awesome

PR #210, opened 2019-11-08
chore(README.md): remove queen's guard emoji and superfluous shields

silasbw:


silasbw:
lol.  i was worried it might be confusing, or have some cultural significance we weren't aware of that might be off putting.

PR #213, opened 2019-11-11
Support mounting existing secrets as files

Pluies:
Hello 👋 
I just realised we need to support mounting a CA file as a secret in this pod to communicate with Vault, which is best done through secrets.
This PR therefore copies the `envVarsFromSecret` construct, but for files.
I also bumped the chart version, but I'm not sure that's the correct way of doing things - please let me know if I need to do it differently!
- Florent

### DIFF
--- a/charts/kubernetes-external-secrets/Chart.yaml
+++ b/charts/kubernetes-external-secrets/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubernetes-external-secrets
-version: 2.1.0
+version: 2.1.1
 appVersion: 2.1.0
 description: Kubernetes External Secrets CustomResourceDefinition
 keywords:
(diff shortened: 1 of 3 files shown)


dschaaff:
I'm currently mounting the vault ca cert using this method. The only piece I'd add is that I had to add the `NODE_EXTRA_CA_CERTS` environment variable pointing to the mounted file.

Pluies:
That's what I'm using it for as well; but I left the env var side out of this PR as there's already support for it on the Helm templating side. This keeps the PR generic enough to be used for any other files as needed, and enables this use-case 👍

keweilu:
We should not change the chart version in the PR. Once the PR is merged, we can release a new chart to include the changes.

keweilu:
Just released a new chart version which should include those changes

PR #217, opened 2019-11-15
fix: update image tag on helm chart to 2.2.0

brucex:


brucex:
@keweilu thanks for the release. looks like the helm chart needs to point to the new version as well.

PR #226, opened 2019-11-21
Add podLabels value in chart

vasrem:
This feature will allow users to specify additional labels to the pods. This is very useful if you for example would like to group components that interact with a specific provider by label. 

Most likely the commit #ed5b199 is not needed. I allow edits from maintainers, so feel free to delete it's not needed, or I can do it myself.


charts/kubernetes-external-secrets/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kubernetes-external-secrets
-version: 2.2.0
+version: 2.2.1
```
keweilu:
```suggestion
version: 2.2.0
```

vasrem:
Could we also have a fresh chart release with this included?

vasrem:
@keweilu ping

keweilu:
@vasrem sorry for the delay, we just released a new version

vasrem:
@keweilu thanks a lot :)

Issue #178, opened 2019-10-17
Enforcing naming conventions for AWS Secrets Manager (and HashiCorp Vault) entries


muenchhausen:
_1. "As cluster operations, I want to force developers to follow namespace based naming conventions on AWS Secrets Manager, 
because secrets are just presented as a single list in the AWS console without any grouping possibilty. There is no other way than naming convention to tidy up and to prevent naming clashes."_

_2. "As an IT security manager, I want to enforce in a simple way, that Team A cannot access secrets of Team B, because AssumeRole is for some setups far too complicated."_

The proposed solution is an enforced naming convention defined as annotation within the namespaces in addition to #143.

**Given:**

As an example AWS Secrets Manager naming convention could look like folder names starting with
deployment stage, followed by a namespace and ending with a secret name:

```
 dev/team-a-namespace/order-service
 dev/team-b-namespace/delivery-service
```

```
kind: Namespace
metadata:
  name: team-a-namespace
  annotations:
    # this annotation checks naming convention for AWS Secrets Manager entries
    secretsmanager.amazonaws.com/naming: dev/team-a-namespace/.*

---
kind: Namespace
metadata:
  name: team-b-namespace
  annotations:
    # this annotation checks naming convention for AWS Secrets Manager entries
    secretsmanager.amazonaws.com/naming: dev/team-b-namespace/.*
```

```
# Team A - valid naming
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  namespace: team-a-namespace
  name: order-service
secretDescriptor:
  backendType: secretsManager
  data:
    # valid naming:
    - key: dev/team-a-namespace/order-service
      name: password
      property: password
      backendType: secretsManager

---
# Team A - invalid naming
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  namespace: team-a-namespace
  name: order-service
secretDescriptor:
  backendType: secretsManager
  data:
    # invalid naming:
    - key: order-service
      name: password
      property: password
      backendType: secretsManager

---
# Team B - no access
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  namespace: team-b-namespace
  name: order-service
secretDescriptor:
  backendType: secretsManager
  data:
    # invalid naming:
    - key: dev/team-a-namespace/order-service
      name: password
      property: password
      backendType: secretsManager

```

**When**
Checking naming convention in the poller module 

[...]

PR #230, opened 2019-11-22
Enforcing naming conventions for AWS Secrets Manager (and HashiCorp Vault) entries

muenchhausen:
See issue #178. 

This patch is tested. It runs smoothly on two clusters since more than 4 weeks.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -24,6 +24,7 @@ const logLevel = process.env.LOG_LEVEL || 'info'
 const pollingDisabled = 'DISABLE_POLLING' in process.env
 
 const rolePermittedAnnotation = process.env.ROLE_PERMITTED_ANNOTATION || 'iam.amazonaws.com/permitted'
+const namingPermittedAnnotation = process.env.NAMING_PERMITTED_ANNOTATION || 'iam.amazonaws.com/naming'
 
 const metricsPort = process.env.METRICS_PORT || 3001
 
@@ -33,6 +34,7 @@ module.exports = {
   pollerIntervalMilliseconds,
   metricsPort,
   rolePermittedAnnotation,
+  namingPermittedAnnotation,
   pollingDisabled,
   logLevel
 }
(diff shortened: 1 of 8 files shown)


muenchhausen:
Documentation is missing for both useages

1. AWS Secrets Manager
2. HashiCorp Vault 

I will add that, if this pull request is of interest.

config/environment.js
```diff
@@ -24,6 +24,7 @@ const logLevel = process.env.LOG_LEVEL || 'info'
 const pollingDisabled = 'DISABLE_POLLING' in process.env
 
 const rolePermittedAnnotation = process.env.ROLE_PERMITTED_ANNOTATION || 'iam.amazonaws.com/permitted'
+const namingPermittedAnnotation = process.env.NAMING_PERMITTED_ANNOTATION || 'iam.amazonaws.com/naming'
```
Flydiverny: Perhaps we want another annotation than this here, but Im not really sure what.
Since its not really related to iam and we its something that can be applied to all backendtypes   
`externalsecrets.kubernetes-client.io/permitted-key-name:` 
@silasbw thoughts?

lib/poller.js
```diff
@@ -187,21 +189,42 @@ class Poller {
...
+
+    const externalData = descriptor.data || descriptor.properties
+    const namingConvention = namespace.metadata.annotations[this._namingPermittedAnnotation]
+
+    if (namingConvention) {
+      externalData.forEach((secretProperty, index) => {
+        const reNaming = new RegExp(namingConvention)
+        if (!reNaming.test(secretProperty.key)) {
+          allowed = false
+          reason = `key name does not match naming convention ${namingConvention}`
```
Flydiverny: might as well return here, now we keep looping and the error message will be for the last key, as opposed to early exit with error for the first key name

config/environment.js
silasbw: sgtm

lib/poller.js
muenchhausen: Yep you are right. I fixed this. Thanks!

config/environment.js
muenchhausen: much better name.

PR #238, opened 2019-11-28
feature/parameterstorebypath

rjmsilveira:
We have a special need which is to simplify the amount of times we have to input any given variable.

The ideia here is that we can fetch all keys inside a given path and export them as secret.
It saves the secrets by using the end of the path as the key instead of having to type them individually.

This setup with the k8s secretRef will mean that we only need to add a variable once in Parameter Store and it will automatically be available in the secrets file and on the pod without having to type it again.

The k8s setup in use is the following:
```
envFrom:
    - secretRef:
      name: ssm-test
```

Please keep in mind that I am not a programmer and you may find here a over complicated approach to this problem. 
The code in this PR is a mix of trial and error with a good amount of documentation and google search.

Let me know if you see this as a good addition to your service.
Thanks

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@ coverage
 Dockerfile
 node_modules
 release.sh
+
--- a/config/index.js
+++ b/config/index.js
@@ -13,6 +13,7 @@ const envConfig = require('./environment')
 const CustomResourceManager = require('../lib/custom-resource-manager')
 const SecretsManagerBackend = require('../lib/backends/secrets-manager-backend')
 const SystemManagerBackend = require('../lib/backends/system-manager-backend')
+const SystemManagerBackendbyPath = require('../lib/backends/system-manager-backend-bypath')
 const VaultBackend = require('../lib/backends/vault-backend')
 
 // Get document, or throw exception on error
@@ -46,12 +47,18 @@ const systemManagerBackend = new SystemManagerBackend({
   assumeRole: awsConfig.assumeRole,
   logger
 })
+const systemManagerBackendbyPath = new SystemManagerBackendbyPath({
+  clientFactory: awsConfig.systemManagerFactory,
+  assumeRole: awsConfig.assumeRole,
+  logger
+})
 const vaultClient = vault({ apiVersion: 'v1', endpoint: envConfig.vaultEndpoint })
 const vaultBackend = new VaultBackend({ client: vaultClient, logger })
 const backends = {
   // when adding a new backend, make sure to change the CRD property too
   secretsManager: secretsManagerBackend,
   systemManager: systemManagerBackend,
+  systemManagerbyPath: systemManagerBackendbyPath,
   vault: vaultBackend
 }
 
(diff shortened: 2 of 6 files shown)


Flydiverny:
Interesting!
So the idea here is to have secrets in ssm with keys for example `/my/secret/password` and `/my/secret/username`

And you would then use an ExternalSecret:
```yaml
kind: ExternalSecret
metadata:
  name: ssm-example-by-path
spec:
  backendType: systemManagerbyPath
  dataFromPath:
    - /my/secret
```

To yield a secret like:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: ssm-example-by-path
type: Opaque
data:
  username: <base64 of /my/secret/username>
  password: <base64 of /my/secret/password>
```

rjmsilveira:
Yes @Flydiverny that is correct.
By doing it like this you end up with a secret file with all the keys on the root of that ssm path and without any extra effort :smile: 
Use that with the k8s secretRef and all the secrets will be available to the pod without having to reference each one of them

Flydiverny:
How would you expect something like:

Keys in secrets manager:
```
/my/secret/app/username
/my/secret/app/password
/my/secret/username
/my/secret/password
```

with:
```
kind: ExternalSecret
metadata:
  name: ssm-example-by-path
spec:
  backendType: systemManagerbyPath
  dataFromPath:
    - /my/secret
```

To generate for result? I guess this is where `Recursive: false` comes in so it would yield the same result as my example above https://github.com/godaddy/kubernetes-external-secrets/pull/238#issuecomment-559316132 ?

rjmsilveira:
Yes its true. `Recursive: false` is there to prevent it from fetching child paths but the ideia is also to have that option (not implemented in this PR) so anyone can decide if they want or not the recursiveness to occur.

Can't think of a good example where that would be of any help apart from organizational point of view.

rjmsilveira:
Hi all. Need your guidance.
What are the next steps I need to do to see this change merged?
Thanks!

bpoland:
This is very helpful, thanks.  I had to modify crd.yaml to accept the new ExternalSecret though.  Also wonder if it would be possible to keep the backendType as systemManager but just accept the dataFromPath underneath it, instead of a new backendType.

bpoland:
@Flydiverny is there any chance of this getting merged?  If it would help I can push up the `crd.yaml` change that I mentioned above.  Right now I'm working off a fork because we need this feature.  Thanks!

martina-if:
@rjmsilveira  Hi! This is a great feature! Will you continue working on this?

rjmsilveira:
Hi @martina-if 
Indeed, It is my wish to continue working on it.
We are currently using this fork in production and its working flawlessly.

@bpoland also made a good point, in which he refers if this could somehow be merged with the current implementation for parameterstore 

Let me see if I can pick it up meanwhile to keep this alive :)

rjmsilveira:
Just letting you know that I will be working on this MR this week :) Expect some news soon (if all goes well :smile: )

rjmsilveira:
@martina-if @bpoland
Here is my refactor. It is indeed more clean than my first approach.
Created a new MR and will close this one.
The code can be found here: https://github.com/external-secrets/kubernetes-external-secrets/pull/603

rjmsilveira:
Duplicate of #603 
Closing this one over the new approach :smile:

PR #245, opened 2019-12-05
Add support for Istio metrics

ejhayes:
Thanks for sharing this! I have this setup with Istio and noticed that gathering metrics was not supported in the helm chart. I added a flag to add support for Istio to scrape these metrics (see: https://istio.io/faq/metrics-and-logs/#prometheus-application-metrics). To enable this, your helm `values.yaml` file should contain:

```
istio:
  metrics:
    enabled: true
```

**NOTE:** I thought this could be accomplished by just adding the annotations to `podAnnotations` value, but was running into some issues when mTLS was configured. In addition to the `podAnnotations` it was also required to add a container port and define a `Service`. Once I added those metrics were able to be collected automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,7 @@ kubernetes-external-secrets exposes the following metrics over a prometheus endp
 | ----------------------------------------- | ------------------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
 | `sync_calls`                              | This metric counts the number of sync calls by backend, secret name and status  | `sync_calls{name="foo",namespace="example",backend="foo",status="success"} 1` |
 
+If you are using Istio you can enable metrics to be gathered automatically by setting the helm value `istio.metrics.enabled` to `true`.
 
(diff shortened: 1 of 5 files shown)


Flydiverny:
I dont see how these things are istio specific so I'd drop the istio naming. 
Also the port could probably always be added, defaulting to 3001 as now.
I find the `metrics.enabled` confusing as metrics will still be generated even if I set this to false. The prometheus annotations could still be added by using `podAnnotations`.
Perhaps we can always create the service as well? Not sure whats best here, if its needed?

ejhayes:
Agree with your input @Flydiverny 

So just expose port 3001 by default and always create the service?

Flydiverny:
@keweilu any thoughts?

keweilu:
yeah, sounds good to me

Flydiverny:
Closing this as I believe #268 solves the same problem

Issue #250, opened 2019-12-12
Feature request for null values from System Manager


manuel-sanchez:
Hi, this is more of a feature request than a bug. SSM doesn't allow to have null values set, however this null values might be useful on the app side. For our app we ended up creating a script that replaces a given value in SSM with null, in our case if the value equals the name of the variable we set the value as null. We could add something similar here if this feature has enough quorum and we can decide a standard. The code change from what I've seen is limited to the system-manager-backend.js

manuel-sanchez:
Here's a PR for this: https://github.com/godaddy/kubernetes-external-secrets/pull/251

Flydiverny:
This feels a bit weird to me, couldn't this be solved by the user? Eg you have a standard in your team that if an ssm value is set to `THIS_IS_NULL` your application could treat it accordingly.
Setting the value to null based on the key name and value matching seems like it could cause a lot of confusion if you are not aware or expecting this behaviour.

manuel-sanchez:
I know this is a weird solution, and definitely parameter store has this shortcoming (not allowing empty values). It doesn't have to be the name of the variable, it could be when the value is `NULL` or some other value. I can move this to be solved on the app side, but if I'm not the only one dealing with this problem it would be nice to have this CDR handle this use case.

manuel-sanchez:
I know this is a weird solution, and definitely parameter store has this shortcoming (not allowing empty values). It doesn't have to be the name of the variable, it could be when the value is `NULL` or some other value. I can move this to be solved on the app side, but if I'm not the only one dealing with this problem it would be nice to have this CDR handle this use case.

PR #251, opened 2019-12-12
Allow null values from System Manager

manuel-sanchez:
If value equals name set the value of the variable to null. This allows to have null values from parameter store


Flydiverny:
I'm closing this as I don't really see this happening. Please continue discussion in #250 if needed

PR #252, opened 2019-12-16
Fix annotation bugs

gerrywastaken:
Before this change, if you tried providing a ServiceAccount annotation
you will get the following error from Helm:
`coalesce.go:199: warning: destination for annotations is a table. Ignoring non-table value <nil>`

This fixes the default value so this issue doesn't happen.

There is also another fix for the annotations in the deployment so that it's more consistent with other areas. Without this, we generate an empty annotations section in the deployment when no podAnnotations are provided.

### DIFF
--- a/charts/kubernetes-external-secrets/values.yaml
+++ b/charts/kubernetes-external-secrets/values.yaml
@@ -33,7 +33,7 @@ serviceAccount:
   # Specifies whether a service account should be created
   create: true
   # Specifies annotations for this service account
-  annotations:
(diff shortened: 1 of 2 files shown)


charts/kubernetes-external-secrets/values.yaml
```diff
@@ -33,7 +33,7 @@ serviceAccount:
   # Specifies whether a service account should be created
```
Flydiverny: 9

Issue #164, opened 2019-09-18
IAM role configuration


res0nat0r:
I have a question about the snippet below from the readme:

> Specify each parameter using the --set key=value[,key=value] argument to helm install. For example,

```
bash helm install --name kubernetes-external-secrets \
--set env.POLLER_INTERVAL_MILLISECONDS='300000' \
--set podAnnotations."iam\.amazonaws\.com/role"='Name-Of-IAM-Role-With-SecretManager-Access' \
charts/kubernetes-external-secrets
```

> Use IAM credentials for Secrets Manager access
> If not running on EKS you will have to use an IAM user (in lieu of a role). 

Does the above assume that the role you are referencing via `Name-Of-IAM-Role-With-SecretManager-Access`, is applied directly to the k8s nodes of the cluster? Just creating an IAM role with only SecretsManagerReadWrite access and applying it to the helm install via the pod annotation isn't sufficient to grant read/write access to any secrets.

If the above code is assuming the arn you're supplying is directly applied to the k8s nodegroup itself, then isn't this a security vulnerability? If the full r/w IAM policy is directly attached to the k8s nodes, then this means anyone running any pod in the k8s cluster can read/write secrets automatically due to the temporary creds they can get via the AWS metadata server at http://169.254.169.254/

reubenavery:
the annotation example provided is for use with [kube2iam](https://github.com/jtblin/kube2iam), an excellent piece of tooling, you should definitely check it out.

mbonig:
> the annotation example provided is for use with [kube2iam](https://github.com/jtblin/kube2iam), an excellent piece of tooling, you should definitely check it out.

if that's the case then it needs to be better documented. The current examples make it sound like it'll 'just work'.

Flydiverny:
To use the secrets manager or system manager backends you have to provide the pod with AWS credentials in some way. How users choose to do that varies, I'd assume that if you have a cluster in place you probably already set this up somehow, if you are just starting heres some examples:
- [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)
- [kiam](https://github.com/uswitch/kiam) or [kube2iam](https://github.com/jtblin/kube2iam)
- Providing AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY to the pod
- node access

[...]

PR #256, opened 2019-12-18
docs: Clarify IAM role usage.

aledeganopix4d:
Clarify that `kubernetes-external-secrets` inherits the IAM role of the node it's running on when situated in AWS, whether its within EKS or otherwise.

Also: thank you for sharing this project, is awesome! :pray:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,11 +50,26 @@ $ helm template -f charts/kubernetes-external-secrets/values.yaml --output-dir .
 
 The generated kubernetes manifests will be in `./output_dir` and can be applied to deploy `kubernetes-external-secrets` to the cluster.
 
-### Use IAM credentials for Secrets Manager access
+### Secrets Manager access
 
-If not running on EKS you will have to use an IAM user (in lieu of a role).
-Set AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY env vars in the session/pod.
-You can use envVarsFromSecret in the helm chart to create these env vars from existing k8s secrets
+For `kubernetes-external-secrets` to be able to retrieve your secrets it will need access to your secret backend.
+
+#### AWS based backends
+
+Access to AWS secrets backends (SSM & secrets manager) can be granted in various ways:
+
+1. Granting your nodes explicit access to your secrets using the [node instance role](https://docs.aws.amazon.com/eks/latest/userguide/worker_node_IAM_role.html) (easy for experimentation, not recommended)
+
+2. [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).
+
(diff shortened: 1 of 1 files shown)


Flydiverny:
While it is true that the aws sdk will fallback to the node role, if available, I'm not sure I would want this written like this. I would not recommend to depend on this role as using the node role to grant permissions will give all pods running on the nodes with this node instance role the possibility to reach your secrets.

Using the node instance role is just one of many ways (see https://github.com/godaddy/kubernetes-external-secrets/issues/164#issuecomment-551989503)

aledeganopix4d:
Thank you for your feedback. I understand your point.
My comment comes from the fact that, when I first used this tool, I was wondering if the role could be used only within EKS (rather than any AWS deployed k8s) and my intention was to point in the right direction whoever might find themselves in my same confusion.

Maybe the readme could be changed as following:
`If not running on EKS you will have to use an IAM user (in lieu of a role).`
->
`If not running on AWS (EKS or otherwise) you will have to use an IAM user (in lieu of a role).`

What do you think?

Flydiverny:
Ah yes, the EKS only part is abit confusing :) 

Perhaps we can write something like:
>For `kubernetes-external-secrets` to be able to retrieve your secrets it will need access to your secret backend. For AWS based backends (SSM & secrets manager) this can be granted in various ways. either by granting your nodes explicit access to your secrets using the [node instance role](https://docs.aws.amazon.com/eks/latest/userguide/worker_node_IAM_role.html) (easy for experimentation, not recommended), by using [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) or using solutions like [kiam](https://github.com/uswitch/kiam) or [kube2iam](https://github.com/jtblin/kube2iam). You can also directly provide AWS access credentials to the pod by env vars.

copy pasta friendly:
```md
For `kubernetes-external-secrets` to be able to retrieve your secrets it will need access to your secret backend. For AWS based backends (SSM & secrets manager) this can be granted in various ways. either by granting your nodes explicit access to your secrets using the [node instance role](https://docs.aws.amazon.com/eks/latest/userguide/worker_node_IAM_role.html) (easy for experimentation, not recommended), by using [IAM roles for service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) or using solutions like [kiam](https://github.com/uswitch/kiam) or [kube2iam](https://github.com/jtblin/kube2iam). You can also directly provide AWS access credentials to the pod by env vars.
```

aledeganopix4d:
> Ah yes, the EKS only part is abit confusing :)
> 

@Flydiverny Thank you very much for your feedback. I've integrated your suggestions in the README, rearranging a little bit the layout.
Let me know what you think about it!

Issue #280, opened 2020-02-11
IAM roles setup help


mo-saeed:
Hi Guys,

Need your help how to install kubernetes-external-secrets on shared k8s cluster:

we have the following 
- 1 shared Kubernetes cluster
- project_A and project_B with 2 AWS accounts and 2 namespaces in the shared k8s cluster

How to achieve with one Kubernetes-external-secret installation that project_A's namespace has only access to AWSACCOUNT_A project_B's namespace has only access to AWSACCOUNT_B

I am using KIAM for Authentication.

Thanks

mo-saeed:
@moolen May be you can help me here :)

mo-saeed:
What I did is the following:
- Create IAM role in kubernetes cluster's AWS account called "ROLE_K8S" trusting kiam-server role.
- Create IAM role in the project A's AWS account called "ROLE_PROJ_A" with assume_role policy 
```
{
    "Version": "2012-10-17",
    "Statement": [{
        "Action": "sts:AssumeRole",
        "Principal": {
            "AWS": [
                "arn:aws:iam::K8S_ACCOUNT_ID:role/ROLE_K8S"
            ]
        },
        "Effect": "Allow"
    }]
}
```
and policy 
```
{
  "Version": "2012-10-17",
  "Statement": [{
    "Effect": "Allow",
    "Action": [
      "secretsmanager:GetResourcePolicy",
      "secretsmanager:GetSecretValue",
      "secretsmanager:DescribeSecret",
      "secretsmanager:ListSecretVersionIds"
    ],
    "Resource": "*"
  }]
}
```
- Install kubernetes-external-secret with pod-annotation `iam.amazonaws.com/role: "arn:aws:iam::K8S_ACCOUNT_ID:role/ROLE_K8S"`
- apply this secret 
```
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  name: hello-service
  namespace: test
spec:
  backendType: secretsManager
  #roleArn: "arn:aws:iam::PROJECT_A_ACCOUNT_ID:role/ROLE_PROJ_A"
  data:
    - key: hello-service/password
      name: password
  # optional: specify a template with any additional markup you would like added to the downstream Secret resource.
  # This template will be deep merged without mutating any existing fields. For example: you cannot override metadata.name.
  template:
    metadata:
      annotations:
        cat: cheese
      labels:
        dog: farfel
````
I got this error in the logs 
```
[...]

PR #282, opened 2020-02-13
WIP: Add support for intermediate role

jaxxstorm:
I've added a WIP tag to this MR because I need feedback about the best way of implementing this.

We've implemented KIAM in a way that means our kiam nodes (our masters, in this case) assume role to an intermediate role. We want to have this capability in kubernetes-external-secrets, so I've done a very hacked together job of supporting this in two ways with this MR for secretsmanager - either via an additional spec field or an environment var.

I'm looking for feedback on a better way of implementing this so that it follows your needs and is testable. I'm a very very novice javascript developer but willing to take feedback and criticism where it's needed. If this isn't functionality you need to support. but looking at #280 seems to indicate other people might have run into this.

Happy to get any feedback

### DIFF
--- a/config/aws-config.js
+++ b/config/aws-config.js
@@ -39,7 +39,7 @@ module.exports = {
     }
     return new AWS.SSM(opts)
   },
-  assumeRole: (assumeRoleOpts) => {
+  assumeRole: (assumeRoleOpts, stsConfig) => {
     const sts = new AWS.STS(stsConfig)
     return new Promise((resolve, reject) => {
       sts.assumeRole(assumeRoleOpts, (err, res) => {
--- a/lib/backends/secrets-manager-backend.js
+++ b/lib/backends/secrets-manager-backend.js
@@ -9,7 +9,7 @@ class SecretsManagerBackend extends KVBackend {
    * @param {Object} client - Client for interacting with Secrets Manager.
    * @param {Object} logger - Logger for logging stuff.
    */
-  constructor ({ clientFactory, assumeRole, logger }) {
+  constructor({ clientFactory, assumeRole, logger }) {
     super({ logger })
     this._client = clientFactory()
     this._clientFactory = clientFactory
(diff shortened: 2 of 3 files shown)


cspargo:
the KIAM feature that this is mirroring is referred to here under "Configuring a Server Role" https://github.com/uswitch/kiam/blob/master/docs/IAM.md

moolen:
@jaxxstorm can you elaborate a bit on your use-case? In the KIAM docs i read:
> Using a separate Server IAM role is desirable in cases where the cluster node instances and IAM roles are replaced frequently.

So the idea is to decouple instance profiles from the software running on it. KES (is this how we abbreviate kubernetes-external secrets, btw?) runs inside the cluster. Most likely already with a specific role provided by KIAM/kube2iam. I dont really see a need for a layer of indirection here.

Q: Do you short-circuit KIAM/kube2iam and use KES with `hostNetwork: true`? That would be a use-case where we would need that layer of indirection.

Q: Do you run KES with `AWS_ACCESS* / AWS_SECRET*` env vars? It would make sense to add a layer of indirection between the IAM user and the roles that consume ssm/sm.

Regarding the implementation i would favor env vars over a spec field (considering it is one of the above use-cases).

cspargo:
for us, it's about managing the role assumption policy at scale. Each of our clusters has it's own IAM role/instance profile for its nodes. Consider if we had 10 roles that we wanted to be able to assume role from any of those cluster roles, in order to access secrets. We would need to list each of those cluster roles in the trust policy for those 10 roles, and in the cluster roles, we need to list each of those 10 roles to allow them to assume them (although we could also address that by sharing a policy). By adding an intermediate role, we only need to list the intermediate role in the trust policy of those 10 roles, and only need to list the intermediate role in the cluster roles. When we add/remove clusters, we only need to update the trust policy in the intermediate role, instead of all 10 roles.

Flydiverny:
Created a new PR to support this feature in #454.

I think you could already support this as well by mounting a `~/.aws/credentials` (or other path specified with `AWS_SHARED_CREDENTIALS_FILE`) file where you set _default_ profile to use the intermediate role and add a source profile which uses the pod credentials. Unless you set `AWS_` or `AMAZON_` env vars with credentials, in which case they would be loaded first as determined by the [default credential chain](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CredentialProviderChain.html#defaultProviders-property)

Issue #291, opened 2020-02-28
dataFrom does not function with permitted-key-name annotation


Strange-Account:
Hi,

Version: 3.1.0
Install: by Helm Chart
Environment: GKE 1.14 cluster

this doesn't function when the namespace has the permitted-key-name annotation:

```
apiVersion: v1
kind: Namespace
metadata:
  name: ns-a
  annotations:
    externalsecrets.kubernetes-client.io/permitted-key-name: "/dev/poc/a/.*"

apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  name: credentials
spec:
  backendType: secretsManager
  dataFrom:
    - /dev/poc/a/credentials
```

The error message is: "ERROR, Cannot read property 'forEach' of undefined"

When I remove the annotation or add a data spec to the external secret it is okay, eg.
```
apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
  name: credentials
spec:
  backendType: secretsManager
  data:
    - key: /dev/poc/a/credentials
      name: password
      property: password
  dataFrom:
    - /dev/poc/a/credentials
```

Strange-Account:
It gets worse:

When i use:

apiVersion: kubernetes-client.io/v1
kind: ExternalSecret
metadata:
name: credentials
spec:
backendType: secretsManager
data:
- key: /dev/poc/a/credentials
name: password
property: password
dataFrom:
- /dev/poc/b/credentials

then the dataFrom part is not checked against the naming convention and is merged.

Flydiverny:
Fixed in #292

PR #292, opened 2020-02-28
Fixed dataFrom property in naming convention verification

Strange-Account:


Flydiverny:
#291

Strange-Account:
What test case(s) do you mean?

Flydiverny:
I think adding a variation or two to the existing test case for naming conventions
https://github.com/godaddy/kubernetes-external-secrets/blob/master/lib/poller.test.js#L757

to make sure it tests the codepath for `dataFrom` usage also and not only the `data` variant, so they work independently, and combined.

PR #302, opened 2020-03-17
Update vault-backend.js

sudarshntn:
As the stringify adds addtional double quotes to the data that is retrieved from the Vault in JSON format we need to parse it to have actual data rather than having a double quotes added data content

### DIFF
--- a/lib/backends/vault-backend.js
+++ b/lib/backends/vault-backend.js
@@ -56,7 +56,7 @@ class VaultBackend extends KVBackend {
     from the Vault in JSON format we need to parse it to have actual data rather than
     having a double quotes added data content.*/
     secretValue = JSON.stringify(secretResponse.data.data)
-    return JSON.secretdataVaule(secretVaule)
+    return JSON.parse(secretVaule)
   }
 }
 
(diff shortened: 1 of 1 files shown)


lib/backends/vault-backend.js
```diff
@@ -52,8 +52,11 @@ class VaultBackend extends KVBackend {
 
     this._logger.debug(`reading secret key ${key} from vault`)
     const secretResponse = await this._client.read(key)
-
-    return JSON.stringify(secretResponse.data.data)
+    /**As the stringify adds addtional double quotes to the data that is retrieved 
+    from the Vault in JSON format we need to parse it to have actual data rather than
+    having a double quotes added data content.*/
+    secretValue = JSON.stringify(secretResponse.data.data)
+    return JSON.secretdataVaule(secretVaule)
```
Flydiverny: This line doesn't look right

sudarshntn: Have fixed it please review it. Thanks

sudarshntn: @Flydiverny : Can you please review I have fixed it.

PR #310, opened 2020-03-25
Remove newlines, spaces, and tabs from status string

davidcorbin:
Vault error messages seem to have newlines and tabs which don't look great in k8s status message. This PR cleans up the status string before sending to k8s.

Before:
Screen Shot 2020-03-24 at 9 22 13 PM

After:
Screen Shot 2020-03-24 at 9 29 21 PM

### DIFF
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -161,6 +161,9 @@ class Poller {
   }
 
   async _updateStatus (status) {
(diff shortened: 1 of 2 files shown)


Flydiverny:
Sounds good to me, can we add a test or two?

davidcorbin:
Sounds good. Will do 👍

PR #311, opened 2020-03-26
fix(azure-registry): handle binary files

lukaspj:
Hi!

# The issue
We had an issue with fetching certificates from the Azure KeyVaults.
Because the certificate were binary files, the Azure KeyVault would base64 encode it, however the @azure/keyvault-secrets library gives no notice or way to detect that the secret is base64 encoded.

Then kubernetes-external-secrets would base64 encode it again, so the binary files were base64 encoded twice, which meant that we had to alter our application to base64 decode the certificates before we could use them.

# This change
Here we add a new `keyOptions` field `isBinary` so that you can explicitly specify that the Azure KeyVault secret is a binary and that no base64 encoding should happen.

Notice that the `isBinary` field does not work together with the `property` field, as the `property` field does JSON-decode and that would definetly not work on a Buffer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,7 +424,7 @@ spec:
   data:
     - key: hello-service/credentials
       name: password
-      isBinary: "true"
+      isBinary: true
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -410,6 +410,24 @@ spec:
       property: value
 ```
 
+Due to the way Azure handles binary files, you need to explicitly let the ExternalSecret know that the secret is binary.
+You can do that with the `isBinary` field on the key. This is necessary for certificates and other secret binary files.
+
+```yml
+apiVersion: kubernetes-client.io/v1
+kind: ExternalSecret
+metadata:
+  name: hello-keyvault-service
+spec:
+  backendType: azureKeyVault
+  keyVaultName: hello-world
+  data:
+    - key: hello-service/credentials
+      name: password
+      isBinary: "true"
+```
+
+
 ## Metrics
 
 kubernetes-external-secrets exposes the following metrics over a prometheus endpoint:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -389,6 +389,24 @@ spec:
...
+kind: ExternalSecret
+metadata:
+  name: hello-keyvault-service
+spec:
+  backendType: azureKeyVault
+  keyVaultName: hello-world
+  data:
+    - key: hello-service/credentials
+      name: password
+      isBinary: "true"
```
Flydiverny: cant this just be `isBinary: true` and we add validation to it here https://github.com/godaddy/kubernetes-external-secrets/blob/master/crd.yaml#L62

lukaspj: I have added it to the CRD with the type "boolean" and corrected the use of the field correspondingly.
I wasn't sure whether it should be added to the CRD since it is Azure specific, and I'm not that familiar with CRDs

crd.yaml
```diff
@@ -68,6 +68,9 @@ spec:
                     type: string
                   property:
                     description: Property to extract if secret in backend is a JSON object
+                  isBinary:
+                    description: Handle the secret as a binary file, used in Azure KeyVault to avoid base64 encoding a binary file as they are already base64 encoded.
```
silasbw: nit: I'd rephrase this a bit to be more prescriptive for casual users.

```
You must set this to true if configuring an item for a binary file stored in Azure KeyVault.  Azure automatically base64 encodes binary files and setting this to true ensures External Secrets does not base64 encode the base64 encoded binary files.
```

lukaspj: Sure thing, I modified the description to what you wrote here

silasbw:
Thanks for the tracking this down and submitting the PR :100: !

PR #338, opened 2020-04-03
Update readme. Azure docs out of order

jdamata:
This block for azure binary files is not under the azure headline. It was bothering me


Flydiverny:
@silasbw looks like CI never ran so cant merge? 😕

PR #344, opened 2020-04-10
Chart optionally installs CRD / CR Manager configurable for more strict clusters

### LINKED ISSUE #249: env.AWS_REGION='us-west-2'  causes failure
when performing the helm install, add `--set env.AWS_REGION=‘us-west-2’` to the command. Apply an external secret and query your ES.  You see an error like `ERROR, Invalid character in header content ["Authorization"]` . I did not encounter the same issue when using the us-east-1 region. Of course us-west-2 is the default so this is not a major issue, however it did lead to confusion amongst our team.

### LINKED ISSUE #300: Add CRD to helm chart
For some reason it's missing from the chart and it's not documented to manually install it.

hfuss:
#### Controller
- Adds `DISABLE_CUSTOM_RESOURCE_MANAGER` env var to the daemon config
- Modifies the `customResourceManager` to use this new flag to determine whether or it not it should attempt to upsert the CRD

#### Chart
- CRD is moved to chart and sourced from `charts/kubernetes-external-secrets/crds/kubernetes-client.io_externalsecrets_crd.yaml`
- Adds `customResourceManagerDisabled` option to the Helm chart. This controls the respective env var, the scope of the cluster role, and determines whether or not to install the CRD as part of the chart.
  - Defaults to `false` to remain backwards compat
- Chart is prepared for future upgrade from `v1` to `v2` for Helm V3 compatibility only if that is desired for future 4.x releases.
- Chart docs updated to assume Helm V3, V2 instructions are still provided
- Sets `AWS_DEFAULT_REGION` env var in the helm chart

#### E2E Tests
- Script adjusted to work with Helm V2/V3 and toggling the custom resource manager
- GH Actions workflow uses a build matrix to test all combinations of Helm V2/V3 and disabling the custom resource manager

### Justification

Closes #300
Also closes #249 

Coupling the CRD creation to a controller requires an [unnecessarily privileged cluster role](https://github.com/godaddy/kubernetes-external-secrets/blob/master/charts/kubernetes-external-secrets/templates/rbac.yaml#L18-L20).

In our OpenShift clusters, our admins would not be willing to grant the controller service account that role. We require the chart itself to install the CRD, and the service account or admin installing the chart would have the `crd:create` access while the controller be more tightly scoped. We have been using this fork in our lab clusters for a few weeks now with no issue.

**This essentially deprecates the custom resource manager so that it can removed entirely in future 4.x releases.**


charts/kubernetes-external-secrets/templates/crd.yaml
```diff
@@ -0,0 +1,118 @@
+{{- if .Values.customResourceManageEnabled | not }}
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: externalsecrets.kubernetes-client.io
+  annotations:
+    app.kubernetes.io/managed-by: helm
```
hfuss, line 8: Not sure how to address the two different copies of the CRD... I'm also leveraging this annotation in the E2E tests which is a bit hacky.

Perhaps in 4.x the custom resource manager can be removed in favor of the chart CRD install?

Flydiverny: So for helm v3 we should probably put the CRD in crds folder, these cant be templated though IIRC.
We could probably source this file in the code as well to avoid copies, and overrwrite `metadata.annotations["app.kubernetes.io/managed-by"]` in the code before upserting it.

For helm v2 this file could be changed to something like

```yaml
{{- if .Values.customResourceManagerDisabled }}
{{- if .Values.crds.create }}
{{- range $path, $bytes := .Files.Glob "crds/*.yaml" }}
{{ $.Files.Get $path }}
---
{{- end }}
{{- end }}
{{- end }}
```

Probably we should also set `DISABLE_CUSTOM_RESOURCE_MANAGER` to `true` by default if running helm v3.

Maybe this is all overkill :D either way we need to support both helm v2 and v3, I'll try to check with silas if he has any opinion on removing the CRD management from the runtime. I think it makes sense to favor the chart installing the CRD.

So perhaps we can go straight for 4.x solution :)

hfuss: Great suggestions for sourcing the CRD and v2 backwards compat, I'll give them a shot.

hfuss: Alright I got the CRD sourced from one location in the repo now, and added your template for helm v2 support. However, I couldn't conditionally set a value depending on whether or not helm v2/v3 was being used so `DISABLE_CUSTOM_RESOURCE_MANAGER` will have to default to `false` for now to remain backwards compat.

The helm docs will need to be updated as a result to include `--skip-crds` and explain how to install the CRD via the chart using v2 and v3.

e2e/run-e2e-suite.sh
```diff
@@ -93,8 +95,52 @@ until kubectl get secret | grep -q ^external-secrets-e2e-token; do \
...
+  --env="AWS_SECRET_ACCESS_KEY=foobar" \
+  --env="AWS_DEFAULT_REGION=us-east-1" \
+  --env="AWS_REGION=us-east-1" \
+  --env="LOCALSTACK_STS_URL=http://sts" \
+  --env="DISABLE_CUSTOM_RESOURCE_MANAGER=true" \
+  --generator=run-pod/v1 \
+  --overrides='{ "apiVersion": "v1", "spec":{"serviceAccountName": "external-secrets-e2e"}}' \
+  e2e --image=external-secrets-e2e:test
+
+echo "Preparing for second round of e2e tests..."
```
hfuss, line 118: This isn't great for E2E testing the custom resource manager disabled/enabled.

I think instead adding flags/env vars to this script and using a build matrix in the workflow file such as:
```yaml
  test-e2e:
    runs-on: ubuntu-latest
    name: E2E
    strategy:
      matrix:
        customResourceManagerDisabled: [true, false]
```

would be cleaner. However, builds will use more resources until the 4.x release where the custom resource manager is removed. What do you think? @Flydiverny @silasbw 

It also begs the question as to whether or not Helm v2 should be tested. The build matrix approach could also be leveraged there to test all combinations of custom resource manager settings and Helm versions.

Though `ubuntu-latest` [does not come with Helm v2](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md) so there would need to be a conditional step for installing Helm v2 in the E2E job.

charts/kubernetes-external-secrets/README.md
```diff
@@ -6,7 +6,7 @@
 
 ```bash
 $ helm repo add external-secrets https://godaddy.github.io/kubernetes-external-secrets/
-$ helm install external-secrets/kubernetes-external-secrets
+$ helm install --skip-crds external-secrets/kubernetes-external-secrets
```
hfuss, line 9: Should it remain:
```bash
$ helm install --name my-release external-secrets/kubernetes-external-secrets
```
?

Which version of Helm are the docs targeting? I cannot think of a way to set the defaults so that the easy install works for both V3 and V2 without one of the versions requiring an additional `--skip-crds` or `--set`.

Flydiverny: I'd say they've been targeting helm v2 but with EOL of v2 around the corner, I'd say it makes sense to target v3 as a default.

hfuss:
Sorry for the delay, finally got around to testing on our lab OpenShift clusters, looks good! This controller is great. I'll rebase off of master and update the docs, and then let's try to get this merged?

Sytten:
If that correct to assume that currently you cant install the service with helm v3? I get an error on startup about not finding /openapi.

hfuss:
@Sytten can you provide more details? Are you saying master currently does not work with helm v3 or this branch specifically? What is erroring on startup?

I have not tried master honestly, but we are using helm v3 in argocd to deploy this version of the chart currently without issue.

Sytten:
@hfuss I created https://github.com/godaddy/kubernetes-external-secrets/issues/392

hfuss:
@Flydiverny @silasbw hey folks, any chance I could get a review this week?

lib/custom-resource-manager.js
```diff
@@ -104,6 +104,20 @@ class CustomResourceManager {
 
     throw new Error(`Unable to update resource ${resourceName}`)
   }
+
+  async manageCrd ({ customResourceManifest }) {
+    this._kubeClient.addCustomResourceDefinition(customResourceManifest)
```
Flydiverny, line 109: Does this always need to be added, perhaps only do this if its not disabled? I'm thinking it would be discovered by `await kubeClient.loadSpec()` in daemon above if CRDs are already in the cluster.

.github/workflows/workflow.yml
```diff
@@ -17,11 +17,23 @@ jobs:
...
+        helmVersion: ['V2', 'V3']
     steps:
       - uses: actions/checkout@v2
       - name: Setup node
         uses: actions/setup-node@v1
         with:
           node-version: 12
+      - uses: azure/setup-helm@v1
+        with:
+          version: v2.16.1
```
Flydiverny, line 32:
```suggestion
          version: v2.16.7
```

Flydiverny:
I can't merge this since the E2E status check is stuck as expected. I assume as thats whats on master, but on this branch its changed (as it should be).

I'll try to poke @silasbw

silasbw:
Thanks @hfuss (and @Flydiverny for pinging me on slack)!  :clinking_glasses:

hfuss:
@Flydiverny yessir! I wrote the changes so that it would be backwards compatible for users who previously installed the chart using Helm V2 and relied on the CR manager. Helm V3 users will relying on  the CR manager will need `--skip-crds` moving forward but the documentation addresses that as best it can. The documentation is also now tailored towards Helm V3 rather than Helm V2. Thanks again for the early feedback as it was very helpful.

And thanks @silasbw for merging this! 🍻

PR #348, opened 2020-04-13
Enable use of AWS STS regional endpoints

Aneurysm9:
The default STS configuration uses the legacy global endpoint, which does not work from private VPCs accessing STS via VPC Endpoints.  This change adds the `region` and `stsRegionalEndpoints` values to the `stsConfig` object used to construct the STS client.  It defaults to "regional" as the value of `stsRegionalEndpoints` but honors the `AWS_STS_ENDPOINT_TYPE` environment variable for users who prefer/require the use of "legacy" endpoints.

### DIFF
--- a/config/aws-config.js
+++ b/config/aws-config.js
@@ -9,7 +9,10 @@ const localstack = process.env.LOCALSTACK || 0
 
 let secretsManagerConfig = {}
 let systemManagerConfig = {}
-let stsConfig = {}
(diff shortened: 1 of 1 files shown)


Flydiverny:
@Aneurysm9 can you rebase or recreate this PR so CI is run, there was an issue with the GH actions workflow not triggering on PRs :)

PR #353, opened 2020-04-21
feat: support isBinary for GCP

### LINKED ISSUE #352: Support `isBinary: true` for GCP secrets manager
This is useful if you have a secret (for example a JSON file) that is already base64 encoded in GCP secrets manager.

With this I can then create a secret that looks like 

```data:
  credentials.json: |
  { ...}
```

which is needed e.g. when writing out GCP SA files.

pmuir:
Fixes #352

### DIFF
--- a/lib/backends/gcp-secrets-manager-backend.js
+++ b/lib/backends/gcp-secrets-manager-backend.js
@@ -17,6 +17,7 @@ class GCPSecretsManagerBackend extends KVBackend {
   /**
    * Get secret property value from GCP Secrets Manager.
    * @param {string} key - Key used to store secret property value in GCP Secrets Manager.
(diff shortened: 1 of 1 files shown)


lib/backends/gcp-secrets-manager-backend.js
```diff
@@ -19,12 +19,16 @@ class GCPSecretsManagerBackend extends KVBackend {
    * @param {string} key - Key used to store secret property value in GCP Secrets Manager.
```
Flydiverny, line 19: can we add some comment 

```suggestion
   * @param {string} key - Key used to store secret property value in GCP Secrets Manager.
   * @param {boolean} keyOptions.isBinary - Is the secret base64 encoded? Set to true to handle as binary.
```

if you have a better suggestion Im game 😅

PR #355, opened 2020-04-22
add support for Alibaba Cloud KMS Secret Manager

yagrxu:
with current version only the followings are supported:
1. 'ACSCurrenit' for version stage is the default value and recommented.
2. AccessKeyId + AccessKeySecret or STS (assume a provided role name using the provided AccessKeyId and AccessKeySecret). ECS_RAM_ROLE and Kube2Ram are not supported yet.
3. secret value as plaintext - parsing secret value as a json object not yet support


yagrxu:
I have tested locally and in Alibaba Cloud K8s cluster as well. Please review and merge

yagrxu:
@silasbw Could you please kindly review the change.

lib/backends/alicloud-secrets-manager-backend.js
```diff
@@ -0,0 +1,52 @@
...
+    const getSecretValueRequest = new GetSecretValueRequest({
+      secretName: key,
+      versionStage: versionStage
+    })
+
+    const client = this._getClient({ specOptions: { roleArn } })
+    const value = await client.getSecretValue(getSecretValueRequest)
+
+    const secret = { value: value.secretData.toString('utf-8') }
+    return JSON.stringify(secret)
```
Flydiverny, line 48: This should probably be something like:
```suggestion
    return value.secretData.toString('utf-8')
```

yagrxu: thank you for the feedback, I accepted and made the update :)

yagrxu:
ready for review @Flydiverny

PR #361, opened 2020-04-26
Improve out-of-the-box compatibility with clusters running PodSecurityPolicy

chadlwilson:
**Background**

We run our EKS clusters with locked-down `PodSecurityPolicy` by default; that is we drop all privileges and change the default EKS policy to one that is essentially unprivileged.

`kubernetes-external-secrets` does not appear to require any particular privileges that are from the set.

This makes two changes to allow the chart/image to work out of the box without configuration on the user side.

1. changes the `Dockerfile` to indicate the user as `uid` 1000 rather than `node`, which is required to indicate K8S that it's not running as root (when running under Docker, at least) - [consistent with the base node images](https://github.com/nodejs/docker-node/blob/master/12/alpine3.11/Dockerfile#L6)
    
    Minimal workaround on `3.2.0` without the change in this PR is a `values.yaml` override of
    ```
    securityContext:
      runAsUser: 1000
    ```
1. changes the defaults to include a `securityContext` that specifies the non root requirement.

This second change is a little more debatable and I am interested in feedback however it seems a sensible idea to give an indication to chart users the actual requirements of `kubernetes-external-secrets` pods so they can craft custom policies if necessary; and understand any possible additional attack surface required for running a security-sensitive tool like this in their cluster (which I would argue is minimal given the pod requirements).

A possible follow-up step would be to allow to specify the **container** `securityContext` rather than just the pod-level; which would allow us to specify
```
securityContext:
  privileged: false
  allowPrivilegeEscalation: false
   capabilities:
    drop: ["ALL"]
  readOnlyRootFilesystem: true
```

This has been tested to work fine and automatically select a relatively unprivileged "secure-by-default" PSP like the below
```yaml
apiVersion: extensions/v1beta1
kind: PodSecurityPolicy
metadata:
  name: unprivileged
spec:
  allowPrivilegeEscalation: false
  defaultAllowPrivilegeEscalation: false
  fsGroup:
    ranges:
    - max: 65535
      min: 1
    rule: MustRunAs
  readOnlyRootFilesystem: true
  requiredDropCapabilities:
  - ALL
  runAsUser:
    rule: MustRunAsNonRoot
  seLinux:
    rule: RunAsAny
  supplementalGroups:
    ranges:
    - max: 65535
      min: 1
    rule: MustRunAs
  volumes:
  - configMap
  - emptyDir
  - projected
  - secret
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,5 +11,6 @@ RUN npm ci --production
 # Copy app to source directory
 COPY . .
 
-USER node
+# Change back to the "node" user; using its UID for PodSecurityPolicy "non-root" compatibility
+USER 1000
 CMD ["npm", "start"]
--- a/charts/kubernetes-external-secrets/values.yaml
+++ b/charts/kubernetes-external-secrets/values.yaml
@@ -62,7 +62,9 @@ fullnameOverride: ""
 podAnnotations: {}
 podLabels: {}
 
-securityContext: {}
+securityContext:
+  runAsNonRoot: true
+  # Required for use of IRSA, see https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html
   # fsGroup: 65534
 
 resources: {}
--- a/charts/kubernetes-external-secrets/README.md
+++ b/charts/kubernetes-external-secrets/README.md
@@ -41,41 +41,41 @@ helm delete my-release
 
 The following table lists the configurable parameters of the `kubernetes-external-secrets` chart and their default values.
 
-| Parameter                            | Description                                                  | Default                                                 |
-| ------------------------------------ | ------------------------------------------------------------ | ------------------------------------------------------- |
-| `env.AWS_REGION`                     | Set AWS_REGION in Deployment Pod                             | `us-west-2`                                             |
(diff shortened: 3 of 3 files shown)


Dockerfile
```diff
@@ -11,5 +11,5 @@ RUN npm ci --production
 # Copy app to source directory
 COPY . .
 
-USER node
+USER 1000
```
Flydiverny, line 14: Maybe a comment to clarify it's the node user if one isn't aware :)

charts/kubernetes-external-secrets/values.yaml
```diff
@@ -62,8 +62,8 @@ fullnameOverride: ""
 podAnnotations: {}
 podLabels: {}
 
-securityContext: {}
-  # fsGroup: 65534
```
Flydiverny, line 66: Perhaps we can keep this, seems like it's useful for people running IRSA, tho I guess it's not really helpful without any introduction anyway 😅

chadlwilson: Thanks, good observation; I've added back with a comment.

Dockerfile
chadlwilson: Good point; added now :)

PR #373, opened 2020-05-06
Handle JSON in GCP Secrets Manager

### LINKED ISSUE #369: JSON objects in GCP Secret Manager
Is it possible to handle JSON objects via GCP manager. For instance if I store the following JSON in a GCP Secret:

```
{
  "foo": "bar"
}
```

I would like it to be handled in a similar way that the docs describe for AWS, i.e. I can make an ExternalSecret like this:

```
kind: ExternalSecret
metadata:
  name: gcp-secrets-manager-example
spec:
  backendType: gcpSecretsManager
  data:
    - key: projects/111122223333/secrets/my-secret/versions/latest
      name: foo
      property: bar
```

However, currently it seems only string secrets are supported as line 50 in [gcp-secrets-manager-backend](https://github.com/godaddy/kubernetes-external-secrets/blob/master/lib/backends/gcp-secrets-manager-backend.js) explicitly puts the secret in another object with a key:
```
const secret = { value: version[0].payload.data.toString('utf8') }
```

Which from my understanding means the property `bar` will not be found in the object in [kv-backend](https://github.com/godaddy/kubernetes-external-secrets/blob/master/lib/backends/kv-backend.js)

```
if (!(property in parsedValue))
```
will be false, as the only allowed property is `value`.

Is there something I'm missing? If not, can this feature be added, please.

### LINKED ISSUE #372: gcp secret
Hello, 

I want to know if they're any reason to store secret in kubernetes (come from GCP SecretsManager) with a JSON like this `{"value": "<my-secret-value"}`, and how is should be exploitable with kubernetes, specially for environment variables imported from secret

Furthermore, secret from AWS SecretsManager seems to be imported as "raw", without this json

Maybe I miss something, but I see anything in the code

Eraac:
Fixes #372 
Fixes #369 

It's a breaking changes, because the content of the secret will change

### DIFF
--- a/lib/backends/gcp-secrets-manager-backend.js
+++ b/lib/backends/gcp-secrets-manager-backend.js
@@ -23,9 +23,10 @@ class GCPSecretsManagerBackend extends KVBackend {
 
   /**
    * Get secret property value from GCP Secrets Manager.
-   * @param {string} key - Key used to store secret property value in Azure Key Vault.
+   * @param {string} key - Key used to store secret property value in GCP Secrets Manager.
    * @param {string} specOptions.projectId - Id of the gcp project, if not passed, this will be fetched from the client auth
    * @param {string} keyOptions.version - If version is passed then fetch that version, else fetch the latest version
+   * @param {string} keyOptions.property - If the backend secret is an object, this is the property name of the value to use.
    * @returns {Promise} Promise object representing secret property value.
    */
   async _get ({ key, keyOptions, specOptions: { projectId } }) {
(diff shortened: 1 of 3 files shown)


lib/backends/gcp-secrets-manager-backend.js
```diff
@@ -47,12 +45,32 @@ class GCPSecretsManagerBackend extends KVBackend {
...
+        ` To use plain text secret remove the 'property: ${keyOptions.property}'`)
+        return
+      }
+
+      if (!(keyOptions.property in parsedValue)) {
+        throw new Error(`Could not find property ${keyOptions.property} in ${key}`)
+      }
+
+      return parsedValue[keyOptions.property]
+    }
```
Flydiverny, line 71: Shouldnt be any need to handle this here, and would probably even cause issues to do so

```suggestion
```

This is handled here:
https://github.com/godaddy/kubernetes-external-secrets/blob/945d1fdb44a17e98fb0281d77fa65c8e0048456d/lib/backends/kv-backend.js#L30-L50

Eraac: Ok, I got it ...

My bad, didn't see the extends KVBackend

PR #389, opened 2020-05-26
[WIP] Add cache for get from backend operations

### LINKED ISSUE #382: Enable secrets cache
Hello,

I would like to raise the idea of having a cache for polling the secrets.
My use case is having the same secret on multiple namespaces.

With the current implementation, the external secrets controller fetch AWS many times, one for each namespace, although they all use the same secrets manager entry.

Adding such a feature will reduce the API calls which will reduce cost.

shaikatz:
Closes #382 

I'm assuming that user understands that there is a cache and if he uses the same key on different ExternalSecret object he will not get a fresh key.

I've also set the cache time to be less then the poll time by default, so you'll usually get a fresh value every poll if the user uses the default values.


lib/object-cache.js
```diff
@@ -0,0 +1,33 @@
...
+    this._cache = new NodeCache({ stdTTL: backendCacheDurationSeconds / 1000 })
+  }
+
+  /**
+   * Set object in the cache
+   */
+  set (object) {
+    const objectHash = hash(object)
+    this._cache.set(objectHash)
+  }
```
Flydiverny, line 22: Doesn't actually set anything in the cache? 
https://github.com/node-cache/node-cache#store-a-key-set

```suggestion
  set (object, value) {
    const objectHash = hash(object)
    this._cache.set(objectHash, value)
  }
```

runningman84:
this seems like a workaround for https://github.com/godaddy/kubernetes-external-secrets/issues/491
Any chance to get the cache implemented?

shaikatz:
I still think that feature could be useful for us as well, I can fix & rebase as long maintainers thinks this is an acceptable feature.

moolen:
hey @shaikatz i think this is useful but also a little scary. Defining the cache key is really hard to get right: it must include fields from the spec because it contains access/account information: For instance a `roleArn`, `region` which is relevant for aws. Others fields are also potentially relevant.
I suggest adding this feature through a feature gate, e.g. `--enable-cache` or via ENV with default=disabled so that a user has to explicitly enable this feature.


I don't think this fully solves #491, it may take a little load off the upstream API if multiple ES reference the same secret.

MeNsaaH:
@Flydiverny How about for a start just refactoring the `_fetchDataValues` function to first group th data before calling the map function. This way, we get to call it once for each key.

https://github.com/external-secrets/kubernetes-external-secrets/blob/3409c66500e25ec2d069585e7d39e9699e790fe1/lib/backends/kv-backend.js#L30


to something like this:
```javascript
  _fetchDataValues ({ data, specOptions }) {
    cleanedData = functionThatCleansData(data)
    # cleanedData
    # [{
    # key: xxx
    # name: xxx
    # properties: [xxx,xxxx,xxx]
    # }]
    return Promise.all(cleanedData.map(async dataItem => {
      const { name, property = null, key, path, ...keyOptions } = dataItem
     
    ...
    # then process the data and return the intended object
```

It might add some computational overhead but it does reduce the number of calls to be made. This could be done in a separate PR tho. I'm not really fluent at javascript, I woulda done it.

PR #409, opened 2020-06-19
feat: allow permitted-key-name to be provided as list

nbendafi-yseop:
While documentation says that `permitted-key-name` can be provided as regular expression
```
kind: Namespace
metadata:
  name: core-namespace
  annotations:
    # annotation key is configurable
    externalsecrets.kubernetes-client.io/permitted-key-name: "/dev/cluster1/core-namespace/.*"
```
it would be nice (clearer ?) if it can support _list_ of regular expressions
```
kind: Namespace
metadata:
  name: core-namespace
  annotations:
    # annotation key is configurable
    externalsecrets.kubernetes-client.io/permitted-key-name:
      - "/dev/cluster1/core-namespace/.*"
      - "/common/.*"
```

### DIFF
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -209,7 +209,12 @@ class Poller {
     // Testing data property
     if (namingConvention && externalData) {
       externalData.forEach((secretProperty, index) => {
-        const reNaming = new RegExp(namingConvention)
+        let reNaming = new RegExp()
+        if (Array.isArray(namingConvention)) {
+          reNaming = new RegExp(namingConvention.join('|'))
+        } else {
+          reNaming = new RegExp(namingConvention)
+        }
         if (!reNaming.test(secretProperty.key)) {
           allowed = false
           reason = `key name ${secretProperty.key} does not match naming convention ${namingConvention}`
@@ -223,8 +228,13 @@ class Poller {
     // Testing DataFrom property
     const externalDataFrom = descriptor.dataFrom
     if (namingConvention && externalDataFrom) {
+      let reNaming = new RegExp()
       externalDataFrom.forEach((secretProperty, index) => {
-        const reNaming = new RegExp(namingConvention)
+        if (Array.isArray(namingConvention)) {
+          reNaming = new RegExp(namingConvention.join('|'))
+        } else {
+          reNaming = new RegExp(namingConvention)
+        }
         if (!reNaming.test(secretProperty)) {
           allowed = false
           reason = `key name ${secretProperty} does not match naming convention ${namingConvention}`
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -205,16 +205,16 @@ class Poller {
 
     const externalData = descriptor.data || descriptor.properties
     const namingConvention = namespace.metadata.annotations[this._namingPermittedAnnotation]
+    let reNaming = new RegExp()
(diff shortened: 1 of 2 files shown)


nbendafi-yseop:
This PR needs #411 for tests to pass

lib/poller.js
```diff
@@ -209,7 +209,12 @@ class Poller {
     // Testing data property
     if (namingConvention && externalData) {
       externalData.forEach((secretProperty, index) => {
-        const reNaming = new RegExp(namingConvention)
+        let reNaming = new RegExp()
+        if (Array.isArray(namingConvention)) {
+          reNaming = new RegExp(namingConvention.join('|'))
+        } else {
+          reNaming = new RegExp(namingConvention)
+        }
```
Flydiverny, line 217: Maybe we can move this out of the loop so its only done once eg moving it after `const namingConvention = namespace.metadata.annotations[this._namingPermittedAnnotation]`, should be fine even if we there's no `namingConvention` set. Keeping the `if` checks the same 

Then we also shouldn't need to repeat the logic for `dataFrom` below

nbendafi-yseop: Yes, totally. I was mislead by the `const reNaming = new RegExp(namingConvention)` in the `forEach` loop.

ronlut:
@nbendafi-yseop Is it possible (k8s perspective) to have a list value for an annotation?
It [looks like it supports only string values](https://cloud.google.com/run/docs/reference/rpc/k8s.io.apimachinery.pkg.apis.meta.v1#k8s.io.apimachinery.pkg.apis.meta.v1.ObjectMeta).

nbendafi-yseop:
@ronlut From my understanding of those statements
>Annotations is an unstructured key value map stored with a resource that may be set by external tools to store and retrieve arbitrary metadata

and 

> The metadata in an annotation can be small or large, structured or unstructured,

yes they are strings but can represent anything that the application, that consumes them, can understand (ex. raw string, base64 encoded, csv,...).

ronlut:
The reason I am asking is I tried to create a namespace with your example, but it doesn't work:
```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Namespace
metadata:
  name: core-namespace
  annotations:
    # annotation key is configurable
    externalsecrets.kubernetes-client.io/permitted-key-name:
      - "/dev/cluster1/core-namespace/.*"
      - "/common/.*"
EOF

error: error validating "STDIN": error validating data: ValidationError(Namespace.metadata.annotations.externalsecrets.kubernetes-client.io/permitted-key-name): invalid type for io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta.annotations: got "array", expected "string"...
```

nbendafi-yseop:
@ronlut 
The creation of this namespace will work if you replace 
```yaml
    externalsecrets.kubernetes-client.io/permitted-key-name:
...
```
by 
```yaml
    externalsecrets.kubernetes-client.io/permitted-key-name: |
...
```
(ie. add a pipe).
This will only avoid k8s validation error, but this might be not properly handled by `external-secret`, since we'll need to _unmarshall_ this string to get the underlying structure (in our case a list represented as a multi-line string). May I suggest you to open another issue and link it with this one, please.

PR #410, opened 2020-06-19
Fix/vault namespace support

rimitchell:
The `requestOptions` nested object was lost during the refactor to support not including headers if the config was not set. It was missed in testing locally because the container used for testing hadn't updated in the k8s cluster.


Flydiverny:
Please rebase or merge master!

Issue #254, opened 2019-12-16
Fargate Support


cabrinha:
Hello, now EKS has Fargate as a supported runtime. In this environment, there are no nodes and no ec2 metadata endpoint.

I tried to run external-secrets in a fargate context and I got errors about failing to connect the the metadata endpoint.

What works well for me is using IAM Roles for Service Accounts, as outlined here: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

However, when the runtime is Fargate, I still get those ec2 metadata endpoint errors.

Is there a way we can use Helm to pass in configuration parameters that the service could be looking for in ec2 metadata?

I'm assuming the issue is here: https://github.com/godaddy/kubernetes-external-secrets/blob/master/config/aws-config.js#L45

And what may be needed here is an option for specifying "Assume Role with Web Identity" instead of just "Assume Role".

Flydiverny:
If its the assume role thing I think the same issue you are experiencing should be reproducible without Fargate or do they limit more there? Haven't had the opportunity to test neither IRSA nor eks+fargate yet.

For reference https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/STS.html#assumeRoleWithWebIdentity-property

cabrinha:
I think that the Assume Role call will check the ec2 metadata endpoint where Assume Role with Web Identity will not: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html#stsapi_comparison

```
Assume Role - IAM user or IAM role with existing temporary security credentials
AssumeRoleWithWebIdentity- Any user; caller must pass a web identity token that indicates authentication from a known identity provider
```

Because Assume Role needs and IAM user or IAM Role, I'm thinking it's hitting that ec2 metadata endpoint to get the user or role making the call.

Assume Role with Web Identity doesn't require any user, just a web identity token.

lukaszbudnik:
Same for me. 

I can get kubernetes-external-secrets working on "old"  EKS, but it fails on Fargate.

I think the fix should be to rely on the default AWS SDK credentials chain provider (I have a go app which uses AWS GO SDK it and works like a charm on Fargate). I'm pretty sure node.js SDK default credentials chain provider supports it too.

For example when I execute this command on the pod I get this:

```
$ kubectl exec external-secrets-kubernetes-external-secrets-58b679b56c-skfq6 env | grep AWS
[...]

PR #414, opened 2020-06-23
Adding Support for EKS Fargate

dalgibbard:
This PR adds in support for Fargate-only EKS Clusters -- so the External Secrets controller pod can run on fargate and still function.
The key difference is that the EC2 metadata API is not available on EKS Fargate instances, and you must instead use a Web Identity.

Summary:
* Call ```assumeRoleWithWebIdentity``` in place of ```AssumeRole``` when ```env.FARGATE=1```
* Read a web token from the service account token mount in the container (usually ```/var/run/secrets/eks.amazonaws.com/serviceaccount/token``` when automounted by K8S)
* Deploy the chart with ```--set env.FARGATE=1``` -- this enables the use of assumeRoleWithWebIdentity instead of assumeRole.
* Deploy the chart with ```--set securityContext.fsGroup=65534``` -- The securityContext.fsGroup sets ownership permissions for the servicetoken; without this, its set to root:root // 0640 and isn't accessible to the app.

PS. I'm a JS noob, so all feedback is welcome!
TODO: Update any readme's/doc's if this change is deemed sensible.

### DIFF
--- a/charts/kubernetes-external-secrets/README.md
+++ b/charts/kubernetes-external-secrets/README.md
@@ -35,6 +35,8 @@ To install the chart with [AWS IAM Roles for Service Accounts](https://docs.aws.
 $ helm install my-release external-secrets/kubernetes-external-secrets --skip-crds --set securityContext.fsGroup=65534 --set serviceAccount.annotations."eks\.amazonaws\.com/role-arn"='arn:aws:iam::111111111111:role/ROLENAME'
 ```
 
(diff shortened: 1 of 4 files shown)


dalgibbard:
Fixes #254

Flydiverny:
Thanks for PR! :) Merged #416 as it seemed like a more generic implementation fix as the same issue exists for IRSA usage in EKS, eg not using the correct assume role. The AWS SDK should pick up and use the web token automatically if set, but we were not switching over to using the correct assume role for those credentials.

So closing this as I don't think we have a need for the `FARGATE` env var. Please reopen if the merged PR doesn't resolve your problem!

dalgibbard:
Hey, Yeap that makes sense, will test and report back :)

PR #425, opened 2020-07-08
fix: Additional quotes added while creating secret

sudarshntn:
As the JSON.stringify adds additional quotes when it reads the json data from the vault and while creating the secret object in K8s it adds additional quotes to avoid it just we need to make sure it is parsed to remove it and have only the actual value used for the secret in K8s. I have made this change in my usage and it is working perfectly fine.

### DIFF
--- a/lib/backends/vault-backend.js
+++ b/lib/backends/vault-backend.js
@@ -53,7 +53,12 @@ class VaultBackend extends KVBackend {
     this._logger.debug(`reading secret key ${key} from vault`)
     const secretResponse = await this._client.read(key)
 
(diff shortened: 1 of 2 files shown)


Flydiverny:
Please provide examples with how you're secrets look and how you are using vault.

If I setup a vault dev server create a secret 
image

And use the vault example 
https://github.com/godaddy/kubernetes-external-secrets/blob/23d511f02901e0f1222b6d89122cc930cffa7939/examples/hello-service-external-secret-vault.yml#L1-L12

The existing solution works perfectly fine?


From what I can see the secrets are always JSON objects and for KV-2 backend the payload is nested in `data.data`

PR #429, opened 2020-07-11
Switch to human-readable log levels

davidcorbin:
Switch from the following log messages:

```
{"level":30,"time":1594499119283,"pid":18,"hostname":"kubernetes-external-secrets-56d5969cd6-gdkb4","msg":"starting poller for asdf/database-credentials"}
```

to 

```
{"level":"info","time":1594499119283,"pid":18,"hostname":"kubernetes-external-secrets-56d5969cd6-gdkb4","msg":"starting poller for asdf/database-credentials"}
```

Many log aggregation tools parse the logs for human readable log levels like "warning" or "error". This makes it easier to send alerts based on those logs, etc.


Flydiverny:
I wouldn't be opposed to changing this, but just straight up changing it would be a breaking change. Would probably be better having it configurable and keeping the current default for existing users.

davidcorbin:
Good point about keeping the legacy format. I've switched the configuration to an env var and added the variable to the helm chart.

PR #448, opened 2020-07-22
added the option to enforce namespace annotations

omerlh:
Namespace annotations are amazing, but there is no way to make them mandatory. Added an env var that when set, makes those annotations mandatory.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -30,7 +30,7 @@ const pollingDisabled = 'DISABLE_POLLING' in process.env
 
 const rolePermittedAnnotation = process.env.ROLE_PERMITTED_ANNOTATION || 'iam.amazonaws.com/permitted'
 const namingPermittedAnnotation = process.env.NAMING_PERMITTED_ANNOTATION || 'externalsecrets.kubernetes-client.io/permitted-key-name'
-const enforceNamespaceAnnotation = process.env.ENFORCE_NAMESPACE_ANNOTATIONS || false
+const enforceNamespaceAnnotation = 'ENFORCE_NAMESPACE_ANNOTATIONS' in process.env || false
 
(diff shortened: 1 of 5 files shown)


lib/poller.js
```diff
@@ -29,6 +29,8 @@ class Poller {
    * @param {Object} customResourceManifest - CRD manifest
    * @param {Object} externalSecret - ExternalSecret manifest.
    * @param {string} rolePermittedAnnotation - namespace annotation that defines which roles can be assumed within this namespace
+   * @param {string} namingPermittedAnnotation - namespace annotation that defines which keys can be assumed within this namespace
+   * @param {string} enforceNamespaceAnnotation - nshould enforce namespace annotations
```
Flydiverny, line 33:
```suggestion
   * @param {string} enforceNamespaceAnnotation - should enforce namespace annotations
```

config/environment.js
```diff
@@ -30,6 +30,7 @@ const pollingDisabled = 'DISABLE_POLLING' in process.env
 
 const rolePermittedAnnotation = process.env.ROLE_PERMITTED_ANNOTATION || 'iam.amazonaws.com/permitted'
 const namingPermittedAnnotation = process.env.NAMING_PERMITTED_ANNOTATION || 'externalsecrets.kubernetes-client.io/permitted-key-name'
+const enforceNamespaceAnnotation = process.env.ENFORCE_NAMESPACE_ANNOTATIONS || false
```
Flydiverny, line 33: to make sure its always a boolean value and not; string | boolean

```suggestion
const enforceNamespaceAnnotation = 'ENFORCE_NAMESPACE_ANNOTATIONS' in process.env || false
```

omerlh:
Thanks!

PR #449, opened 2020-07-23
feat: support isBinary for Vault

ilkh:


lib/backends/vault-backend.js
```diff
@@ -55,10 +57,16 @@ class VaultBackend extends KVBackend {
     const secretResponse = await this._client.read(key)
 
     if (kvVersion === 1) {
+      if (keyOptions && keyOptions.isBinary) {
+        secretResponse.data[property] = Buffer.from(secretResponse.data[property], 'base64').toString('utf8')
```
Flydiverny, line 61: This feels kind of scary 😄 I guess its fine to turn it into a string but feels like we have an underlying problem in the code forcing us to write weird code :)
Main issue is probably that the KV backend expects us to return a string in `_get` forcing us to serialise here when we already have an object.

pkrishnath:
Exactly what i am looking for .. Is there any way to store a base64 encoded value in vault?

lib/backends/vault-backend.js
Flydiverny, line 61: Would it be enough to not do the `toString('utf8')` here, just converting the base64 string to a buffer. later on this buffer will be base64 encoded again as the base class won't mess with buffers. This should put the original value from vault into the secret without it ending up double base64 encoded.

https://github.com/godaddy/kubernetes-external-secrets/blob/c3c27bc427c7129e4e98b5b10e4c5dde146c422d/lib/backends/kv-backend.js#L91-L93

https://github.com/godaddy/kubernetes-external-secrets/blob/c3c27bc427c7129e4e98b5b10e4c5dde146c422d/lib/backends/kv-backend.js#L130

Flydiverny (changes requested):
Needs rebase and comments to be addressed. Should try returning a buffer from vault-backends get instead of a binary string.

Flydiverny:
Closing this in favor of #585

PR #472, opened 2020-08-17
feat: Cluster level default settings for Hashicorp Vault

sukratkashyap:
This PR enables global default `Vault Role` and `Vault Mount Point` setting for k8s authentication to Hashicorp Vault. 

**Reason:**
When a user tries to deploy an app to multiple different cluster which uses the same vault. The `External Secret` yaml file has differences since, `vaultRole` and `vaultMountPoint` could be different for different clusters. This leads to multiple `External Secret` yaml file. One for each cluster. Whereas this could be specified during deployment of External Secret helm chart. 

Since, a cluster will most probably will have one `vaultMountPoint`. It is better to have a way to deploy it at the config level. Leaving the user not to worry about role and mount point and focus on the values retrieved from the vault. Since, it is a default value. This will be backwards compatible with the previous method as it can be overwritten with the value in the `External Secret`  manifest file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,11 +227,11 @@ by default an `ExternalSecret` may access arbitrary keys from the backend e.g.
       name: password
 ```
 
-An enforced naming convention helps to keep the structure tidy and limits the access according 
-to your naming schema. 
+An enforced naming convention helps to keep the structure tidy and limits the access according
+to your naming schema.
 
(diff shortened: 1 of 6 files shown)


megakid:
Note that my new PR would be compatible with this change. #488

Flydiverny (approved):
LGTM

Issue #354, opened 2020-04-22
AWS secretes manager field to provide a region


amithapa:
```yaml
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  name: hello-service
spec:
  backendType: secretsManager
  # optional: specify role to assume when retrieving the data
  roleArn: arn:aws:iam::123456789012:role/test-role
  data:
    - key: hello-service/password
      name: password
  # optional: specify a template with any additional markup you would like added to the downstream Secret resource.
  template:
    metadata:
      annotations:
        cat: cheese
      labels:
        dog: farfel
```
My kubernetes nodes are running in different regions and I have created aws secrets specific to that region. 
E.g. External secrets aws_region is set us-west-1 and I am trying to fetch secrets from us-east-1.

Kindly provide a feature for defining region while creating ExternalSecrets.

argais:
This would be extremely useful in multi-region scenarios, for both secrets manager and parameter store.
An IAM role can easily give access to parameters/secrets in multiple regions, but we have no way to specify that right now. 
Either a region field, or being able to specify the full arn to the secret/parameter would be useful.

moolen:
We have a use-case for this, too. I'll pick this up.

moolen:
#475 was merged, this is done :heavy_check_mark:.

PR #475, opened 2020-09-01
feat(aws): add region support to ssm and sm

moolen:
This PR adds multi-region support for ssm and sm #354

I feel like we should (somehow) cache & reuse the ssm/sm `client` object so we don't assume a role in each `_get` call. E.g. by using the hash of the `ExternalSecret` or `${namespace}/${name}/${metadata.generation}` as cache key for the client object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,8 @@ spec:
   backendType: secretsManager
   # optional: specify role to assume when retrieving the data
   roleArn: arn:aws:iam::123456789012:role/test-role
+  # optional: specify region
+  region: us-east-1
   data:
     - key: hello-service/credentials
       name: password
@@ -315,6 +317,8 @@ spec:
   backendType: secretsManager
   # optional: specify role to assume when retrieving the data
   roleArn: arn:aws:iam::123456789012:role/test-role
+  # optional: specify region
+  region: us-east-1
   dataFrom:
     - hello-service/credentials
 ```
@@ -330,6 +334,8 @@ spec:
   backendType: secretsManager
   # optional: specify role to assume when retrieving the data
   roleArn: arn:aws:iam::123456789012:role/test-role
+  # optional: specify region
+  region: us-east-1
   dataFrom:
     - hello-service/credentials
   data:
--- a/e2e/tests/secrets-manager.test.js
+++ b/e2e/tests/secrets-manager.test.js
@@ -124,6 +124,70 @@ describe('secretsmanager', async () => {
     expect(secret.body.type).to.equal('kubernetes.io/tls')
   })
 
+  it('should pull existing secret from secretsmanager in the correct region', async () => {
+    const smEU = awsConfig.secretsManagerFactory({
+      region: 'eu-west-1'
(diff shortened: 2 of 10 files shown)


moolen:
@silasbw do you have some time to look into this?

Flydiverny:
Yo! I merged #454 which caused a bunch of conflicts here.
I think caching client is probably good, we just merged a PR for vault doing something similar.

moolen:
Thanks for the update @Flydiverny, i rebased my changes.

Flydiverny (commented):
Changes looks good to me, but one question. If I pass a region, is this for only specifying the region for the secret? Which region is the IAM role in? Currently I think the IAM role would be assumed using the default region. But I could be wrong about that. What would be the expected behavior?

moolen:
Thanks for the feedback! True, documentation is lacking. The readme is already really long, we should consider using something to generate static pages and serve them using github pages (i'll open a issue for that). Let me try to outline the mechanics:

Take this slightly modified [example from the AWS docs](https://docs.aws.amazon.com/mediaconnect/latest/ug/iam-policy-examples-asm-secrets.html):

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "secretsmanager:GetResourcePolicy",
                "secretsmanager:GetSecretValue",
                "secretsmanager:DescribeSecret",
                "secretsmanager:ListSecretVersionIds"
            ],
            "Resource": [
                "arn:aws:secretsmanager:us-west-2:111122223333:secret:aes128-1a2b3c",
                "arn:aws:secretsmanager:us-central-1:111122223333:secret:aes192-4D5e6F",
                "arn:aws:secretsmanager:us-east-1:111122223333:secret:aes256-7g8H9i"
            ]
        }
    ]
}
```
IAM roles are global, the attached policies grant access to certain resources in certain regions (or `*`).
When we construct the secretsManager/SSM client through `_clientFactory` it needs to know from what region to fetch the secret or parameter from. This is what the `region` property tries to solve.

If it is not defined explicitly it falls back to use ([see docs](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-credentials-node.html)):
* environment variables `AWS_REGION` or `AWS_DEFAULT_REGION` (that's what's most likely in use - see [helm chart](https://github.com/godaddy/kubernetes-external-secrets/blob/master/charts/kubernetes-external-secrets/values.yaml#L14-L16) )
* shared credentials file `~/.aws/config`
* instance metadata service API ([docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html)) if it runs on EC2

So, essentially, if the region property is set it will take precedence over the above chain.

Flydiverny:
Yes, we should definitely set up some docs. Was planning to try setting something up with [docusaurus](https://docusaurus.io/) but didn't get around to it.

Think we can merge this as is then.

Flydiverny (approved):

Issue #474, opened 2020-08-27
[vault backend] support for multitenancy


raffaelespazzoli:
If I understand correctly, when using Vault as the backend, the external-secrets pod service account is used to retrieved the secrets from vault via the kubernetes authentication method.
This means that every namespace shares the same credentials as far a Vault access goes. 
This obviously does not play well in a multitenant environment in which each namespace is assign to a different tenant and strict isolation between tenants must exist.
Can you confirm that this is the case? 
If it is indeed the case what workarounds do exist?
Here are of a couple proposals:
1. the external-secrets controller can be scoped to watch only one namespace. One controller per namespace will be created. This has the undesirable drawback that a bunch of pods (mostly idle) will need to be created.
2. The external-secrets controller uses a service account of the namespace in which the external secret is defined to perform the kubernetes authentication to vault. The service account can be specified as an additional parameter and it could default to the default service account. This would be the ideal soltion.

devth:
I came here with the same concern.

In our cluster, each namespace gets its own `ServiceAccount` that provides it access to specific paths in vault using a Vault Policy. Number 2 in the above comment would be a nice solution.

Issue #47, opened 2019-04-16
Consider merging with aws-secret-operator


max-lobur:
https://github.com/mumoshu/aws-secret-operator is a very similar concept, consider merging two projects.

silasbw:
Doh -- just when you think you've scoured the internet.  Thanks @max-lobur.  Yes, we I agree, we should definitely consider merging.

lhotrifork:
There's also this :) 
https://github.com/ContainerSolutions/externalsecret-operator

silasbw:
....and https://github.com/cmattoon/aws-ssm :)

JacopoDaeli:
Ok ok, but how should we deal with it? What would be the right approach here for considering to merge?

reegnz:
I believe the right approach would be to reach out to these projects and ask them if they want to unify their efforts in a single project.
Maybe open github issues in each of those projects that links to this issue to start a discussion?

riccardomc:
Hi there, and thank you for reaching out!

I am one of the contributors of https://github.com/ContainerSolutions/externalsecret-operator and I think that we would definitely benefit from aligning on a common CRD spec.

We deliberately wanted to keep it simple, so our CRs currently look like this:

```yaml
apiVersion: externalsecret-operator.container-solutions.com/v1alpha1
kind: ExternalSecret
metadata:
  name: example-externalsecret
spec:
  Key: example-externalsecret-key
  Backend: asm-example
```

But it might not be enough for the others?

reegnz:
I played around with the idea of doing something similar to what you get with StorageClass, PersistentVolumeClaim and PersistentVolume.

I wrote the idea down in a gist to start a discussion about that (maybe it's overcomplicated):
https://gist.github.com/reegnz/acc1f97fe3e78746394103e9c96d9c76

riccardomc:
I like the idea of using annotations.

We had a very similar implementation to support multiple backends on a single operator deployment: https://github.com/ContainerSolutions/externalsecret-operator/issues/6 but didn't consider using annotations.

We decided to simplify this to one operator deployment, one backend. I believe it is easier to reason about and demand namespace isolation to RBACs. See:

- https://github.com/ContainerSolutions/externalsecret-operator/issues/10
- https://github.com/ContainerSolutions/externalsecret-operator/pull/18

A few questions:
- How would you handle multiple deployments of the same operator with distinct backends?
[...]

Issue #132, opened 2019-07-23
Feature: multiple backend in the same external secret


Sytten:
Since the project is expanding, it should be possible to have secrets coming from various places in the same externalsecret. That would break the current YAML though so it's quite a major change.

Flydiverny:
Is there really a use case for this though? I mean is there any case where it would end up being an actual problem to create an additional external secret? Seems like a good way to handle it, I'd expect most uses to stick to the same backend type throughout unless migrating from one secret storage to another.

Sytten:
I don't exactly remember the use case since I don't work at the same company anymore, but I feel you might want some information in SM (databases creds for example to use rotation), but most of the stuff in SSM since you have to pay for it.

DWSR:
This is something we'd be very interested in as we currently use HashiVault and SSM Parameter Store for secrets. Using multiple backends enables an easier transition between secret backends and also enables us to avoid duplicating secrets amongst the various backends.

Sytten:
Not stale

Sytten:
Not stale

Flydiverny:
This will not be added to KES.
Neither has it been included in the CRD spec we came up with for ESO. I think the best thing now would be to make a feature request on the rewrite instead.

Issue #81, opened 2019-06-03
Init Container / Volume frontend


silasbw:
Implement a merge-patch in the [admission webhook server](https://github.com/godaddy/kubernetes-external-secrets/issues/79) to inject an Init Container that implements the ["memory" API](https://github.com/godaddy/kubernetes-external-secrets/issues/78).  The Init Container [example](https://github.com/godaddy/kubernetes-external-secrets/pull/77) might provide some useful implementation details.

- [ ] End-to-end implementation of "memory" frontend
- [ ] Update documentation for using the memory frontend

After completing this work, kubernetes-external-secrets should have full support for a memory front end (see https://github.com/godaddy/kubernetes-external-secrets/issues/46).

PR #477, opened 2020-09-01
Standardize CRD Spec

moolen:
## Request for comments

The idea is to use this draft PR for discussing **only the CRD Spec** mentioned in #47.

:exclamation:  Technology discussions are off-topic

CHANGELOG:
* rename `backend`  to `store`
* remove inline `backend` from `ExternalSecret`
* rename `externalSecretClassName` to `className`
* integrate `dataFrom` into spec

TODO:
* [ ] how would one sync secrets within kubernetes (probably similar to `#474` w/ service accounts)?

### DIFF
--- a/keps/20200901-standard-crd.md
+++ b/keps/20200901-standard-crd.md
@@ -247,10 +247,9 @@ spec:
     kind: SecretStore # ClusterSecretStore
     name: my-store
   target:
-    secret:
-      name: my-secret
-      template:
-        type: kubernetes.io/TLS
+    name: my-secret
+    template:
+      type: kubernetes.io/TLS
   data:
     tls.crt:
       key: /corp.org/dev/certs/ingress
--- a/keps/20200901-standard-crd.md
+++ b/keps/20200901-standard-crd.md
@@ -246,7 +246,7 @@ spec:
   storeRef:
     kind: SecretStore # ClusterSecretStore
     name: my-store
-  frontend:
+  target:
     secret:
       name: my-secret
       template:
--- a/keps/20200901-standard-crd.md
+++ b/keps/20200901-standard-crd.md
@@ -195,7 +195,7 @@ apiVerson: kes.io/v1alpha1
 kind: SecretBackend # or GlobalSecretBackend
 metadata:
   name: vault
-  namespace: example-ns # TODO: namespaced?
+  namespace: example-ns
 spec:
   vault:
     server: "https://vault.example.com"
--- a/keps/20200901-standard-crd.md
+++ b/keps/20200901-standard-crd.md
@@ -1,0 +1,163 @@
+```yaml
+---
+title: Standardize ExternalSecret CRD
+authors: all of us
+creation-date: 2020-09-01
+status: draft
+---
+```
+
+# Standardize ExternalSecret CRD
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Terminology](#terminology)
+- [Use-Cases](#use-cases)
+- [Proposal](#proposal)
+  - [API](#api)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This is a proposal to standardize the `ExternalSecret` CRD in an combined effort through all projects that deal with syncing external secrets. This proposal aims to do find the _common denominator_ for all users of `ExternalSecrets`.
+
+## Motivation
+
+There are a lot of different projects in the wild that essentially do the same thing: sync secrets with Kubernetes. The idea is to unify efforts into a single project that serves the needs of all users in this problem space.
+
+As a starting point i would like to define a **common denominator** for a CustomResourceDefinition that serves all known use-cases. This CRD should follow the standard alpha -> beta -> GA feature process.
+
+Once the CRD API is defined we can move on with more delicate discussions about technology, organization and processes.
+
+### Goals
+
+- Define a alpha CRD
+- Fully document the Spec and use-cases
+
+### Non-Goals
+
+This KEP proposes the CRD Spec and documents the use-cases, not the choice of technology or migration path towards implementing the CRD.
+
+## Terminology
+
+* Kubernetes External Secrets `KES`: A Application that runs a control loop which syncs secrets
+* KES `instance`: A single entity that runs a control loop.
+* ExternalSecret `ES`: A CustomResource that declares which secrets should be synced
+* Backend: A **source** for secrets. The Backend is external to the cluster. E.g.: Alibabacloud SecretsManager, AWS SystemsManager, Azure KeyVault...
+* Frontend: A **sink** for the synced secrets. Usually a `ConfigMap` or `Secret`
+* Secret: credentials that act as a key to sensitive information
+
+## Use-Cases
+
+The following use cases are taken from
+
+1. AS a KES user i want to run multiple KES instances per cluster (e.g. one KES instance per DEV/PROD)
+2. AS a KES user i want to integrate **multiple backends** from a **single KES instance** (e.g. dev namespace has access only to dev secrets)
+3. AS a KES user i want to control the sink for the secrets (aka frontend: store secret as `kind=ConfigMap` or `kind=Secret`)
+4. AS a KES user i want to fetch **from multiple** Backends and store the secrets **in a single** Frontend
+5. AS a KES operator i want to limit the access to certain backends or subresources (e.g. having one central KES instance that handles all ES - similar to `iam.amazonaws.com/permitted` annotation per namespace)
+
+## Proposal
+
+### API
+
+The `ExternalSecret` CustomResourceDefinition is **namespaced**. It defines the following:
+1. source for the secret (backend)
+2. sink for the secret (fronted)
+3. and a mapping to translate the keys
+
+```yaml
+apiVersion: kes.io/v1alpha1
+kind: ExternalSecret
+metadata: {...}
+
+spec:
+
+  # optional.
+  # used to select the correct KES instance (think: ingress.ingressClassName)
+  # There is no need for a indirection (e.g. having a extra resource like kind=IngressClass)
+  # the KES controller is instantiated with a specific class name
+  # and filters ES based on this property
+  externalSecretClassName: "dev"
+
+  # This is the "simple" version to specify a single backend which will be used by all keys
+  # we can have multiple backends per ES (they can be declared per key)
+  # Each backend does need a configuration (e.g. AWS IAM Roles, AWS Region, GCP project-id, )
+  backend:
+    type: secretsManager
+    # optional additional config
+    secretsManager:
+      region: eu-central-1
+      accessKeyID: AKIAIOSFODNN7EXAMPLE
+      roleARN: arn:aws:iam::YYYYYYYYYYYY:role/kes-full-access
+
+  # there can only be one frontend per ES
+  # this is the "thing" that is created by KES.
+  # conceptually speaking a frontend is just a thing that can hold k/v pairs
+  frontend:
+    secret:
+      # do we need a api version here? who handles upgrades?
+      apiVersion: v1
+      name: my-secret
+      template:
+        type: kubernetes.io/dockerconfigjson # or TLS...
+        metadata:
+          annotations: {}
+          labels: {}
+    # of course only one frontend should be defined
+    configMap:
+      # do we need a api version here? who handles upgrades?
+      apiVersion: v1 #
+      name: my-configmap
+      template:
+        metadata:
+          labels:
+            app: foo
+
+  data:
+    # EXAMPLE 1: simple mapping
+    # one key in a backend may hold multiple values
+    # we need a way to map the values to the frontend
+    # it is the responsibility of the backend implementation to know how to extract a value
+  - key: /corp.org/dev/certs/ingress
+    property: pubcert
+    name: tls.crt
+  - key: /corp.org/dev/certs/ingress
+    property: privkey
+    name: tls.key
+
+  # EXAMPLE 2: multiple backends per ES
+  # we also need a way to fetch secrets from multiple backend
+  # thus, we need a way to define a backend per data key
+  - key: /rds/database-secret
+    name: database
+    backend:
+      type: SecretsManager
+      secretsManager:
+        roleARN: arn:aws:iam::YYYYYYYYYYYY:role/kes-ssm-access
+  - key: /users/my-db-user
+    name: username
+    backend:
+      type: SystemsManager
+      systemsManager:
+        roleARN: arn:aws:iam::XXXXXXXXXXXX:role/kes-sm-access
+
+# status holds the timestamp and status of last last sync
+status:
+  lastSync: 2020-09-01T18:19:17.263Z # ISO 8601 date string
+  status: success # or failure
+
+```
+
+This API makes the options more explicit rather than having annotations.
+
+
+## Alternatives
+
+###
--- a/keps/20200901-standard-crd.md
+++ b/keps/20200901-standard-crd.md
@@ -117,13 +117,6 @@ metadata: {...}
 
 spec:
 
-  # optional.
-  # used to select the correct KES instance (think: ingress.ingressClassName)
-  # There is no need for a indirection (e.g. having a extra resource like kind=IngressClass)
-  # the KES controller is instantiated with a specific class name
-  # and filters ES based on this property
-  className: "dev"
-
   # the amount of time before the values will be read again from the store
   refreshInterval: "1h"
 
@@ -198,6 +191,12 @@ metadata:
   namespace: example-ns
 spec:
 
+  # optional.
+  # used to select the correct KES controller (think: ingress.ingressClassName)
+  # The KES controller is instantiated with a specific controller name
+  # and filters ES based on this property
+  controller: "dev"
+
   store:
     # store implementation
     vault:
@@ -236,8 +235,6 @@ spec:
 ```
 
 Workflow in a KES instance:
-1. A user creates a CRD with a certain `spec.externalSecretClassName`
-2. A controller picks up the `ExternalSecret` if it matches the `className`
-3. a) It fetches the `SecretStore` or `GlobalSecretStore` defined in `spec.storeRef` and applies it. This does also apply to `spec.data[].storeRef`
-
-## Alternatives
+1. A user creates a Store with a certain `spec.controller`
+2. A controller picks up the `ExternalSecret` if it matches the `controller` field
+3. The controller fetches the secret from the provider and stores it as kind=Secret in the same namespace as ES
--- a/keps/20200901-standard-crd.md
+++ b/keps/20200901-standard-crd.md
@@ -49,13 +49,13 @@ This KEP proposes the CRD Spec and documents the use-cases, not the choice of te
 * Kubernetes External Secrets `KES`: A Application that runs a control loop which syncs secrets
 * KES `instance`: A single entity that runs a control loop.
 * ExternalSecret `ES`: A CustomResource that declares which secrets should be synced
-* Backend: A **source** for secrets. The Backend is external to the cluster. E.g.: Alibabacloud SecretsManager, AWS SystemsManager, Azure KeyVault...
+* Store: Is a **source** for secrets. The Store is external to KES. It can be a hosted service like Alibabacloud SecretsManager, AWS SystemsManager, Azure KeyVault...
 * Frontend: A **sink** for the synced secrets. Usually a `ConfigMap` or `Secret`
 * Secret: credentials that act as a key to sensitive information
 
 ## Use-Cases
-* one global KES instance that manages ES in **all namespaces**, which gives access to **all backends**, with ACL
-* multiple global KES instances, each manages access to a single or multiple backends (e.g.: shard by stage or team...)
+* one global KES instance that manages ES in **all namespaces**, which gives access to **all stores**, with ACL
+* multiple global KES instances, each manages access to a single or multiple stores (e.g.: shard by stage or team...)
 * one KES per namespace (a user manages his/her own KES instance)
 
 ### User definitions
@@ -65,14 +65,14 @@ This KEP proposes the CRD Spec and documents the use-cases, not the choice of te
 ### User Stories
 From that we can derive the following requirements or user-stories:
 1. AS a KES operator i want to run multiple KES instances per cluster (e.g. one KES instance per DEV/PROD)
-2. AS a KES operator or user i want to integrate **multiple backends** from a **single KES instance** (e.g. dev namespace has access only to dev secrets)
+2. AS a KES operator or user i want to integrate **multiple stores** from a **single KES instance** (e.g. dev namespace has access only to dev secrets)
 3. AS a KES user i want to control the sink for the secrets (aka frontend: store secret as `kind=ConfigMap` or `kind=Secret`)
-4. AS a KES user i want to fetch **from multiple** Backends and store the secrets **in a single** Frontend
-5. AS a KES operator i want to limit the access to certain backends or subresources (e.g. having one central KES instance that handles all ES - similar to `iam.amazonaws.com/permitted` annotation per namespace)
+4. AS a KES user i want to fetch **from multiple** stores and store the secrets **in a single** Frontend
+5. AS a KES operator i want to limit the access to certain stores or subresources (e.g. having one central KES instance that handles all ES - similar to `iam.amazonaws.com/permitted` annotation per namespace)
 
-### Backends
+### Stores
 
-These backends are relevant:
+These stores are relevant:
 * AWS Secure Systems Manager Parameter Store
 * AWS Secrets Manager
 * Hashicorp Vault
@@ -86,7 +86,7 @@ These backends are relevant:
 
 * Kind=Secret
 * Kind=ConfigMap
-* *potentially* we could sync backend to backend
+* *potentially* we could sync store to store
 
 ## Proposal
 
@@ -95,7 +95,7 @@ These backends are relevant:
 ### External Secret
 
 The `ExternalSecret` CustomResourceDefinition is **namespaced**. It defines the following:
-1. source for the secret (backend)
+1. source for the secret (store)
 2. sink for the secret (fronted)
 3. and a mapping to translate the keys
 
@@ -106,26 +106,8 @@ metadata: {...}
 
 spec:
 
-  # optional.
-  # used to select the correct KES instance (think: ingress.ingressClassName)
-  # There is no need for a indirection (e.g. having a extra resource like kind=IngressClass)
-  # the KES controller is instantiated with a specific class name
-  # and filters ES based on this property
-  externalSecretClassName: "dev"
-
-  # the amount of time before the values will be read again from the backen
+  # the amount of time before the values will be read again from the store
   refreshInterval: "1h"
-
-  # This is the "simple" version to specify a single backend which will be used by all keys
-  # we can have multiple backends per ES (they can be declared per key)
-  # Each backend does need a configuration (e.g. AWS IAM Roles, AWS Region, GCP project-id, )
-  backend:
-    type: secretsManager
-    # optional additional config
-    secretsManager:
-      region: eu-central-1
-      accessKeyID: AKIAIOSFODNN7EXAMPLE
-      roleARN: arn:aws:iam::YYYYYYYYYYYY:role/kes-full-access
 
   # there can only be one frontend per ES
   # this is the "thing" that is created by KES.
@@ -151,10 +133,11 @@ spec:
             app: foo
 
   data:
+
     # EXAMPLE 1: simple mapping
-    # one key in a backend may hold multiple values
+    # one key from a store may hold multiple values
     # we need a way to map the values to the frontend
-    # it is the responsibility of the backend implementation to know how to extract a value
+    # it is the responsibility of the store implementation to know how to extract a value
   - key: /corp.org/dev/certs/ingress
     property: pubcert
     name: tls.crt
@@ -162,21 +145,25 @@ spec:
     property: privkey
     name: tls.key
 
-  # EXAMPLE 2: multiple backends per ES
-  # we also need a way to fetch secrets from multiple backend
-  # thus, we need a way to define a backend per data key
+  # EXAMPLE 2: multiple stores per ES
+  # we also need a way to fetch secrets from multiple stores
+  # thus, we need a way to define a store per data key
   - key: /rds/database-secret
     name: database
-    backend:
-      type: SecretsManager
-      secretsManager:
-        roleARN: arn:aws:iam::YYYYYYYYYYYY:role/kes-ssm-access
+    storeRef:
+      name: foo
   - key: /users/my-db-user
     name: username
-    backend:
-      type: SystemsManager
-      systemsManager:
-        roleARN: arn:aws:iam::XXXXXXXXXXXX:role/kes-sm-access
+    storeRef:
+      name: bar
+
+  # used to fetch all properties from a secret
+  # properties are merged in specified order
+  dataFrom:
+  - key: /user/all-creds
+    # optional: reference store
+    storeRef:
+      name: fuu
 
 # status holds the timestamp and status of last last sync
 status:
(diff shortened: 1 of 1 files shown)


keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,163 @@
...
+
+### API
+
+The `ExternalSecret` CustomResourceDefinition is **namespaced**. It defines the following:
+1. source for the secret (backend)
+2. sink for the secret (fronted)
+3. and a mapping to translate the keys
+
+```yaml
+apiVersion: kes.io/v1alpha1
```
moolen, line 76: not sure if this is a good idea maybe we should mount this under `k8s.io`, like `external-secrets.k8s.io`

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,240 @@
...
+
+### User definitions
+* `operator :=` i manage one or multiple `KES` instances
+* `user :=` i only create `ES`, KES is managed by someone else
+
+### User Stories
+From that we can derive the following requirements or user-stories:
+1. AS a KES operator i want to run multiple KES instances per cluster (e.g. one KES instance per DEV/PROD)
+2. AS a KES operator or user i want to integrate **multiple stores** from a **single KES instance** (e.g. dev namespace has access only to dev secrets)
+3. AS a KES user i want to control the sink for the secrets (aka frontend: store secret as `kind=ConfigMap` or `kind=Secret`)
```
mcavoyk, line 69: Is storing the fetched ExternalSecrets into a ConfigMap something that should be solved? This seems likely to end up with secrets material stored in ConfigMaps potentially leading to other privilege escalations.

I do think there is a user story around needing to provide an application with a configuration, and having a secret be a field in the configuration. This could be solved with another solution like allowing templating within the ExternalSecret or referencing an existing ConfigMap with the configuration but without the secret.

DirtyCajunRice:
I feel it is a disservice to generalize "many projects in the wild" and then almost line for line copy the itscontained/secret-manger crd with no attribution to the source.

Flydiverny:
> I feel it is a disservice to generalize "many projects in the wild" and then almost line for line copy the itscontained/secret-manger crd with no attribution to the source.

I think its fair to add a list of links to existing work which were used for inspiration.

keps/20200901-standard-crd.md
Flydiverny: Config maps as a frontend seems funky, to me, in the context of secrets. Tho I could see there being use cases for it. There shouldn't really be any issue creating a configmap instead of a secret.

But perhaps in that case the actual resource should be different eg `ExternalConfig` instead of `ExternalSecret`? Not sure if there's added value in doing the separation, but I feel like it would be more explicit and allows more RBAC configuration

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,240 @@
...
+* AWS Secure Systems Manager Parameter Store
+* AWS Secrets Manager
+* Hashicorp Vault
+* Azure Key Vault
+* Alibaba Cloud KMS Secret Manager
+* Google Cloud Platform Secret Manager
+* Kubernetes (see #422)
+* noop (see #476)
+
+### Frontends
```
Flydiverny, line 85: Are the file frontends (https://github.com/godaddy/kubernetes-external-secrets/issues/81) interesting to bring up here? I guess thats more of a `noop` frontend tho. So you can have a side-car fetch populate a volume with the secret data without it ending up in a k8s secret.

keps/20200901-standard-crd.md
Flydiverny: `external-secrets.k8s.io` sounds nice 😏 unless someone gets unhappy with the use of *.k8s.io

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,240 @@
...
+  # and filters ES based on this property
+  className: "dev"
+
+  # the amount of time before the values will be read again from the store
+  refreshInterval: "1h"
+
+  # there can only be one frontend per ES
+  # this is the "thing" that is created by KES.
+  # conceptually speaking a frontend is just a thing that can hold k/v pairs
+  frontend:
```
Flydiverny, line 122: Something that's come up in the past is potentially merging/patching with an existing secret.
Is that a use case that we find interesting? I suppose you could do something like install an existing helm chart, it includes a secret. You want to append or overwrite a specific key and the external secret operator could potentially do that. It would be a bit odd with partial ownership of the resource, as that would potentially limit some things.

So for example you could then have a `targetRef` similar to how `scaleTargetRef` works for HPAs?

Flydiverny (commented):
Some random thoughts 🙃

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,240 @@
...
+  # we also need a way to fetch secrets from multiple stores
+  # thus, we need a way to define a store per data key
+  - key: /rds/database-secret
+    name: database
+    storeRef:
+      name: foo
+  - key: /users/my-db-user
+    name: username
+    storeRef:
+      name: bar
```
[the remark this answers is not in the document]
mcavoyk: I like this syntax, `data` could be a map of key references while `dataFrom` could be a list of key references.

```yaml
data:
  tls.crt:
    key: /corp.org/dev/certs/ingress
    property: pubcert
  tls.key:
    key: /corp.org/dev/certs/ingress
    property: privkey

dataFrom:
- key: /rds/database-secret
```

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,240 @@
...
+  - key: /corp.org/dev/certs/ingress
+    property: privkey
+    name: tls.key
+
+  # EXAMPLE 2: multiple stores per ES
+  # we also need a way to fetch secrets from multiple stores
+  # thus, we need a way to define a store per data key
+  - key: /rds/database-secret
+    name: database
+    storeRef:
```
mcavoyk, line 160: Having an ExternalSecret able to reference multiple stores complicates the ability to what secrets have been created from a specific store. This field seems better at the spec level rather than in each data reference.

Container are already able to mount any number of secrets, if data from multiple stores is needed in a single secret for another reason, the design around being able to merge ExternalSecrets data into an existing secret seems a bit clearer for this use case - https://github.com/godaddy/kubernetes-external-secrets/pull/477/files#r487474442,

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,240 @@
...
+metadata: {...}
+
+spec:
+
+  # optional.
+  # used to select the correct KES instance (think: ingress.ingressClassName)
+  # There is no need for a indirection (e.g. having a extra resource like kind=IngressClass)
+  # the KES controller is instantiated with a specific class name
+  # and filters ES based on this property
+  className: "dev"
```
mcavoyk, line 114: `className` doesn't fit on the ExternalSecret resource because a controller wouldn't be scoped to a set of ExternalSecrets, but rather scoped to a set of SecretStores.

This would fit better as either `className` or `controller` field on the SecretStore type, allowing the controller to only operate on SecretStores (and ExternalSecrets which reference those stores) that the controller is scoped to. This could make using multiple controllers with specific pod access to their stores more simple to setup.

As an example, the AWS ExternalSecret controller with a specific IAM permissions given to the pod would only fetch values for the AWS Secret Manager SecretStore as both the controller and the SecretStore are scoped to the controller class "aws-secret". The Vault ExternalSecret with Kubernetes service account mounted for permissions would only fetch values for Vault SecretStore as both are scoped to the controller class "vault-secret"

This type of SecretStore scoping really only work if ExternalSecrets were limited to referencing a single store, as per my other comment. Otherwise no single controller would have the privileges necessary to create the resulting secret resource.

keps/20200901-standard-crd.md
mcavoyk: I'm not sure I see the use case of either having the ExternalSecret create a ConfigMap, or having an ExternalConfig resource. Config should be part of application specification, within the helm chart or kubernetes manifests. This is safe to check in to git or other systems. ExternalSecrets, like ConfigMaps could also be safely stored in git and the secret is able to be fetched at deploy time,

Take something like Prometheus AlertManager config, which may need Slack/Pagerduty or other apikeys but is unable to source them from environment variables or other means (ref https://github.com/prometheus/alertmanager/issues/504), this still shouldn't have a frontend of a ConfigMap because it contains secrets.

I'd rather see ExternalSecret take the non-secret configuration and merge with the fetched external secrets, either directly or with a reference to another secret/configmap.

```yaml
spec:
  ... 
  data:
  - slack:
      key: /corp.org/slack
      property: api_url

# Secret is fetched and then merged and templated
  template:
    data:
      alertmanager.yaml: |
        receivers:
        - name: slack
          slack_configs:
          - channel: #alerts
            api_url: {{ .data.slack | quote }}
```

```yaml
spec:
  ... 
  data:
  - slack:
      key: /corp.org/slack
      property: api_url

# Secret is fetched and then merged and templated within the referenced configMap data
  templateFrom:
  - configMap:
      name: alertmanager
      items:
       - key: alertmanager.yaml
```

keps/20200901-standard-crd.md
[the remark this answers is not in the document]
DirtyCajunRice: +1 for the data block, but im not sure that the store reference belongs in data

keps/20200901-standard-crd.md
Flydiverny: Yeah, I feel having different stores for different keys adds complexity for sure. For KES there was some feature requests for this before in https://github.com/godaddy/kubernetes-external-secrets/issues/132 
Main use case there seemed to be migration from one secret store to another. Which I'm not sure if that's enough of a reason to add the complexity? If one really is migrating from one store to another wouldn't you try to at least migrate all the keys used in the same secret. 

Using multiple external secrets to target the same secret could indeed also be an alternative if we explore that further.

keps/20200901-standard-crd.md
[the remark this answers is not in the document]
moolen: > I like this syntax, data could be a map of key references while dataFrom could be a list of key references.

:+1: @mcavoyk This is just for maintaining the correct order, right? 

> im not sure that the store reference belongs in data

@dirtycajunrice We need to merge multiple secrets from $CLOUDPROVIDER into a single secret. E.g. Alertmanager needs multiple API Keys.

keps/20200901-standard-crd.md
moolen: I feel like this is out of scope for now (this is a really big feature)

keps/20200901-standard-crd.md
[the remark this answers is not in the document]
DirtyCajunRice: @moolen 
> +1 @mcavoyk This is just for maintaining the correct order, right?
`dataFrom` was designed to be a way to use all properties from a key as is, without having to specify them individually.
> @DirtyCajunRice We need to merge multiple secrets from $CLOUDPROVIDER into a single secret. E.g. Alertmanager needs multiple API Keys.
Ill reply to that one there :smile:

keps/20200901-standard-crd.md
DirtyCajunRice: Ill give both sides of my thoughts: 

For: This (partial object ownership) is completely within the scope of kubernetes design choices [reference - Server Side Apply](https://kubernetes.io/docs/reference/using-api/api-concepts/#server-side-apply).

Against: It is a very large ask for a very niche gain that - ideally - would be better time spent fixing the upstream charts/repositories where the issue stems from.

keps/20200901-standard-crd.md
moolen: I don't have a requirement for ConfigMaps. Lets remove it from this spec. If someone needs it later on we can add it.

keps/20200901-standard-crd.md
mcavoyk: Individual secret keys can be mounted as volumes as pods or referenced as environment variables already. So SecretStore migration already seems possible using two secrets and changing the pod spec to use the new secret.

keps/20200901-standard-crd.md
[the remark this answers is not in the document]
Flydiverny: >👍 @mcavoyk This is just for maintaining the correct order, right?

> dataFrom was designed to be a way to use all properties from a key as is, without having to specify them individually.

Yepp, in KES `dataFrom` is expected to be a JSON payload. Each property from the object will be set as a field in the generated secrets data. If there are multiple sources with the same resulting field the last source in the `dataFrom` that defines that field will be used. And then if the same field is set in the external secrets `data` that will override any `dataFrom` field.

For `data` the ordering doesn't matter as the name would be unique anyway. So it would just be silly to duplicate to use the same 😄

keps/20200901-standard-crd.md
moolen: Thanks for the discussion! I'm soon gonna push commits that (hopefully) integrate your thoughts. Changelog:
* removed `ConfigMap`
* added `templateFrom` 

I'm resolving this conversation for now to tidy up the discussion. Feel free to un-resolve if there's something missing or wrong.

keps/20200901-standard-crd.md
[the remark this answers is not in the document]
moolen: same as above: thank you for your input! I integrated your changes into the spec. Feel free to un-resolve if something is missing or wrong.
Changelog:
* using data object syntax in favor of array
* add meta info for implementation specifics

keps/20200901-standard-crd.md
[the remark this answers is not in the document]
moolen: `ConfigMaps` are out. resolving this, too.

moolen:
> [...] almost line for line copy the itscontained/secret-manger crd with no attribution to the source.

I added all projects directly involved with the subject. I don't know if it's just me, but this sounds very polemic to me. If you take a close look at the commit history you will see that the origins of the spec lie in KES. From there i tried to improve by eliminating the tedious elements, repetitions and added a bunch of items from my wishlist. Eventually i went through the itscontained spec and aligned the spec in regards to the naming (`Backend` vs. `Store` for example - which was initially a optional thing). @DirtyCajunRice I never wanted to sell "your stuff" as mine or even had the intention to do so.

I pushed a bunch of changes that resulted from the discussion so far. I see a bunch of existing topics and would like to propose the following:

* move the `injection` (via #81 [here](https://github.com/godaddy/kubernetes-external-secrets/pull/477#discussion_r487587449) and [here](https://github.com/godaddy/kubernetes-external-secrets/pull/477#discussion_r487473812) i think?) topic to the "backlog" (maybe a section in the spec?)

For the ongoing discussions (let me try to summarize this) we have:
* one backend vs. multiple backends per `ExternalSecret`
  * related: the controller `className` or `controller` field should be in `ExternalSecretSpec` vs. `StoreSpec`
* merging/patching existing secrets with `targetRef` [here](https://github.com/godaddy/kubernetes-external-secrets/pull/477#discussion_r487474442)

keps/20200901-standard-crd.md
moolen: > migration

I think it is simply impossible to support all facets of a migration. It depends on the application, the context of the environment, the data sources. One thing that should be supported tho is secret rotation within a Secret Provider (think AWS SSM's `AWSCURRENT/AWSPREVIOUS` semantics) but that's provider specific and not part of the CRD.

 > Having an ExternalSecret able to reference multiple stores complicates

I think we should remove the support for multiple-providers from the Spec. It's simpler for the controller to handle, (probably) simpler for a ES user to reason about. right now there is no need for it and we might add it later on.

DirtyCajunRice:
@moolen We are here collaborating together - OSS is built off of a foundation of _improving_ existing ideas. I just think it all goes down the toilet when the chain of idea attribution is broken (why we put kes and cert-manager front and center as attribution on our readme). No harm no foul 🤓. 

Injection does not fit into MVP but was stated just so it was kept in sight. Agree with backlog

keps/20200901-standard-crd.md
mcavoyk: Some ideas around this: 
```yaml
apiVersion: external-secrets.k8s.io/v1alpha1
kind: ExternalSecret
metadata:
  name: my-secret
  ...

spec:
  # target defines the resource which will be created by the ExternalSecret
  target:
    # The secret name of the resource
    # defaults to .metadata.name of the ExternalSecret, immutable
    name: other-secret

    # Enum with values: 'Owner', 'Merge', or 'None'
    # Default value of 'Owner'
    # Owner creates the secret and sets .metadata.ownerReferences of the resource
    # Merge does not create the secret, but merges in the data fields to the secret
    # None does not create a secret (future use with injector)
    creationPolicy: 'Merge'

    template:
     ... 

  data:
    ...
```

I think this should be scoped around creating/managing `Kind: Secret`, `apiVersion: v1` resources for now, so not sure we need a full targetReference.

knelasevero:
Hey! So, to add to the discussion, also prefer one backend per ES, basically because of the same reasons already stated here. 

If we backlog  'merging/patching existing secrets' what does that mean for the CRD? We foresee it somehow?

Right now at Container Solutions we are moving our ES solution forward in the direction of the CRD draft at least in the obviously good things like secret refreshing and multiple key values per secret (things that were on our backlog anyways). Even if we decide to move to a solution built from scratch later on, I think it is good so I have some people with momentum to move to the new effort. Anyway, we want to help wherever needed!

keps/20200901-standard-crd.md
moolen: ES is now allowed to reference only a single store. I added a `controller` field to the `SecretStore` type.

keps/20200901-standard-crd.md
moolen: I think this is a really interesting use-case. I added that to the spec;

Flydiverny:
I like the latest updates 👍

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,254 @@
...
+    # store implementation
+    vault:
+      server: "https://vault.example.com"
+      path: secret/data
+      auth:
+        kubernetes:
+          path: kubernetes
+          role: example-role
+          secretRef:
+            name: vault-secret
```
RichiCoder1, line 217: Should `SecretStore` have a status? W/ secret stores implementing some sort of liveliness check so you can validate a `SecretStore` is healthy before adding secrets?

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,254 @@
...
+  # used to fetch all properties from a secret.
+  # if multiple dataFrom are specified, secrets are merged in the specified order
+  dataFrom:
+  - key: /user/all-creds
+
+# status holds the timestamp and status of last last sync
+status:
+  lastSync: 2020-09-01T18:19:17.263Z # ISO 8601 date string
+  status: success # or failure
+
```
RichiCoder1, line 182: A `message` field with the stringified error would be great for diagnostics.

keps/20200901-standard-crd.md
mcavoyk: I think this would be useful but slightly challenging. Some SecretStore types may not have permissions to do inspection, for instance if Vault role isn't granted[ lookup-self](https://www.vaultproject.io/api-docs/auth/token#lookup-a-token-self) permissions then the SecretStore may report failed liveness but still have permission to fetch ExternalSecrets.

I think my recommendation would be to implement this with best effort and documenting the potential permissions store backends may need to support the liveness check.

keps/20200901-standard-crd.md
mcavoyk: Agree, any reason not to go with a condition array block here for readiness?

moolen: Sure, conditions would fit in here nicely!

What about:
```yaml
status:
  # represents the current phase of secret sync:
  # * Pending | ES created, controller did not (yet) sync the ES or other dependencies are missing (e.g. secret store or configmap template)
  # * Syncing | ES is being actively synced according to spec
  # * Failing | Secret can not be synced, this might require user intervention
  # * Failed  | ES can not be synced right now and will not able to
  # * Completed | ES was synced successfully (one-time use only)
  phase: Syncing
  lastSyncTime: 2020-09-01T18:19:17.263Z # ISO 8601 date string
  failureReason: "..."
  failureMessage: "..."
  
  conditions:
  - type: InSync
    status: "True" # False if last sync was not successful
    reason: "SecretSynced"
    message: "Secret was synced"
    lastTransitionTime: "2019-08-12T12:33:02Z"
    lastSyncTime: "2020-09-23T16:27:53Z"
```

keps/20200901-standard-crd.md
moolen: Agree, determining the `status` of an SecretStore is challenging. I think it would be great to have `Events` on the `SecretStore` CRD this way it is easier to observe if a single ES or multiple ES are having problems.

A basic `is ready` / `is running` could look like this:

```yaml

status:
  # * Pending: prerequisites missing / e.g. referenced secret containing credentials
  # * Running: all dependencies are met
  phase: Running
  conditions:
  - type: Ready
    status: "False"
    reason: "ErrorConfig"
    message: "Unable to assume role arn:xxxx"
    lastTransitionTime: "2019-08-12T12:33:02Z"
    lastHeartbeatTime: "2020-09-23T16:27:53Z"
```

keps/20200901-standard-crd.md
mcavoyk: Looks good, a few comments
* `lastHeartbeatTime` doesnt need to be used here, that is generally only for conditions which need explicit heart beats
* `failureReason`, `failureMessage` already fit within the condition, those would better fit with events in this case rather than explicitly in the base status.

keps/20200901-standard-crd.md
moolen: I stumbled upon this slack thread in [sig-architecture](https://kubernetes.slack.com/archives/C5P3FE08M/p1600968242008600) which points to the [api-review process](https://github.com/kubernetes/community/blob/master/sig-architecture/api-review-process.md#voluntary). From my understanding we really shouldn't use `k8s.io`. If we plan to move the effort towards a Kubernetes SIG we should plan to use `x-k8s.io`.

keps/20200901-standard-crd.md
moolen: (this was added to the backlog)

keps/20200901-standard-crd.md
moolen: Regarding `lastHeartbeatTime`: i think this is the case here. The `InSync` condition should be updated with every sync. We should consider renaming this to `lastSyncTime`.

`failure*` as attributes are duplicates, true. Ill remove them.

keps/20200901-standard-crd.md
moolen: i think this is resolved. if not pls reopen.

Flydiverny:
In #413 there's a PR to recreate the target resource / secret if its deleted. Is this something we should cover here? I think it makes sense and adds a behaviour similar to deployments and pods.

moolen:
@Flydiverny i added a behavior section for the `ExternalSecret` resource. If i understand correctly, the underlying issue there is rate limiting of the AWS parameter store - this is a really important thing to note!

Flydiverny:
> @Flydiverny i added a behavior section for the `ExternalSecret` resource. If i understand correctly, the underlying issue there is rate limiting of the AWS parameter store - this is a really important thing to note!

Yepp, thats one of the underlying issues people have. Sometimes people also want to just trigger a refresh manually, especially interesting when polling is disabled. I guess in the end it boils down to not being able to have "live" updates due to API rate limit/cost or implementation limits.

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,277 @@
...
+  # optional.
+  # used to select the correct KES controller (think: ingress.ingressClassName)
+  # The KES controller is instantiated with a specific controller name
+  # and filters ES based on this property
+  controller: "dev"
+
+  store:
+    # store implementation
+    vault:
```
knelasevero, line 222: We've being playing with the idea of the secret store here at CS, and we think that would be simpler to avoid having different keys for each store type.

So, instead of having:

```
vault:
      server: "https://vault.example.com"
      path: path/on/vault/store
      auth:
     .
     .
     .
# 
## or
# 
# gsm:
#     projectID: some-project-id
#     auth:
#     .
#     .
#     .
```

We could just have:

```
  store:
    type: vault
    parameters:
      server: "https://vault.example.com"
      path: path/on/vault/store
    auth: 
    .
    .
    .
  #
  ## or
  #
  # store:
  #   type: gsm
  #   parameters:
  #     projectID: external-secrets-operator
  #   auth: 
  #   .
  #   .
  #   .
```

Then different implementations, that ask for different parameters, can just pass different things. And we just take care in the implementation to be open to receive anything passed under parameters. I think it is also easier to support new store types and avoid having to change the CRD to have more keys for stores. What do you think?

moolen: I had that thought, too. If we can guarantee that 1) the CRD is validated properly and 2) we are not limited to k/v pairs in the `parameters` field we should do that. But i think `kubebuilder/controller-tools` doesn't support that yet. I value CRD validation a lot. just my two cents.

knelasevero: I don't think we will be limited by specific k/v pairs here.  @iamcaleberic What do you think about CRD validation here?

iamcaleberic: @moolen the controller runtime does have a `runtime.RawExtension` that we are using so as to not be limited to key:value pairs https://godoc.org/k8s.io/apimachinery/pkg/runtime#RawExtension.

moolen: @iamcaleberic as discussed with @knelasevero: It's not suuper important to me, let's do it!

I've added it to the CRD spec.

moolen:
First of all i'd like to thank all of you for your feedback, comments and effort you brought into this. It's good to see that here are a lot of people that want to drive this forward. Thank you!

I think this CRD spec is good enough to be implemented in a PoC. I'd like to leave this PR open for now, i don't think there's a lot of value having this CRD Spec in the repo :sweat_smile:. Let's continue the discussion about merging efforts in #47.

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,273 @@
...
+metadata: {...}
+
+spec:
+
+  # the amount of time before the values will be read again from the store
+  # may be set to zero to fetch and create it once
+  refreshInterval: "1h"
+
+  # there can only be one target per ES
+  target:
```
iamcaleberic, line 125: @moolen What is the difference between `target` in `ExternalSecret` and `frontend` in `SecretStore`

jonatasbaldin:
Hey peeps, I'm implementing the CRDs to bootstrap the project and got a question regarding the `SecretStore` authentication.

### The Container Solutions project uses:
```yaml
...
kind: SecretStore
...
auth: 
  secretRef: 
    name: externalsecret-operator-credentials-asm
 ```

Where, inside the `externalsecret-operator-credentials-asm` Secret, we have the following data:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: credentials-asm
  labels:
    type: asm
type: Opaque
stringData:
  credentials.json: |-
    {
      "accessKeyID": "",
      "secretAccessKey": "",
      "sessionToken": ""
    }
```

Each Store backend unmarshals the JSON data (with the hardcoded `credentials.json`), verifiy the fields and authenticates (or fails to).

### The iscontained project uses:
A specific struct for each Store backend, with the required fields, like:
```golang
type AWSAuth struct {
	AccessKeyID *smmeta.SecretKeySelector `json:"accessKeyID,omitempty"`
	SecretAccessKey *smmeta.SecretKeySelector `json:"secretAccessKey,omitempty"`
	Role *smmeta.SecretKeySelector `json:"role,omitempty"`
}
```

This is possible because there's no `type` field in the CRD, each Store backend has its own struct under `SecretStore.spec`, like `SecretStore.spec.aws`.

---

I'd like to have inputs on what you think is the best approach.

I believe the **iscontained** approach could be better:
- The end user doesn't need to understand why there's a `Secret` with a `credentials.json` key inside
- The CRD would tell exactly which information is required for authentication, without needing to unmarshal/validate a different JSON content for each Store backend
- On top of this, I think using the `SecretStore.spec.aws` approach could be better than having a `SecretStore.spec.type`, since the authentication fields are different for each `type`. This requires a change to the CRD every time a new Store backend is added, but the code would have better validation and be more explicit

Any thoughts?

Flydiverny:
Wouldn't it be possible even with the type field to have an type specific auth struct?
Losing the added `credentials.json` sounds nice tho :) 
The was some discussion on types vs different keys here https://github.com/external-secrets/kubernetes-external-secrets/pull/477#discussion_r508320641

jonatasbaldin:
@Flydiverny 

With the defined CRD spec:
```yaml
...
spec:
  store:
    type: vault
    parameters: # provider specific k/v pairs
      server: "https://vault.example.com"
      path: path/on/vault/store
    auth: {} # provider specific k/v pairs
```

We need to have the following:
- `type` would be a `string` field
- `parameters` would be a `runtime.RawExtension` to allow different parameters for different Store backends, not allowing any validation on the CRD level
- `auth` would be a Secret reference with _different fields_ for each Store backend (like accessKey/secretKey (AWS), projectID (GCP) etc)

The main problem becomes the lack of validation on the CRD level and the need to implement logic on each backend to decode the Secret reference and find the specific key/value pairs for its authentication.

---

If we implement like `iscontained` is doing:
```yaml
...
spec:
  aws:
    region: ...

    authSecretRef:
      accessKeyID:
        key: access-key
        name: aws-key
      secretAccessKey:
        key: secret-key
        name: aws-key
```

We have:
- Validation at the CRD level for each Store backend, so we can easily prevent the creation of a resource without the required keys
- Easily link the authentication data to a Secret reference
- Keep each Store backend code specific and more explicit

---

Even though we would need to add new types inside the CRD each type a new Store backend is implemented, I believe the overall architecture/code base would be more explicit and clean than requiring each backend to decode an interface when authenticating.

What do you think? @Flydiverny @moolen @knelasevero

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,272 @@
...
+This KEP proposes the CRD Spec and documents the use-cases, not the choice of technology or migration path towards implementing the CRD.
+
+We do not want to sync secrets into a `ConfigMap`.
+
+## Terminology
+
+* Kubernetes External Secrets `KES`: A Application that runs a control loop which syncs secrets
+* KES `instance`: A single entity that runs a control loop.
+* ExternalSecret `ES`: A CustomResource that declares which secrets should be synced
+* Store: Is a **source** for secrets. The Store is external to KES. It can be a hosted service like Alibabacloud SecretsManager, AWS SystemsManager, Azure KeyVault...
```
jonatasbaldin, line 63: I suggest changing the word `Store` to `Provider` when referring to external "Providers", like AWS SM.

I believe this would avoid confusion with the `SecretStore` CRD.

Any thoughts?

jonatasbaldin: An user would say:
- "We need to setup a new `SecretStore` with the `AWSSM` provider", instead of
- "We need to setup a new `SecretStore` with the `AWSSM` store"

moolen: totally makes sense!

moolen:
@jonatasbaldin 
> If we implement like iscontained is doing: [...]
i totally agree!

I would like to move this CRD Spec document to the (yet to be created) v2 Repository. This way it should be easier to propose changes.
Ack?

jonatasbaldin:
@moolen yeah, let's do it!

We can work on the `external-secrets` _repository_ for now, I think it would be easier than creating a new empty branch here. I should push the bootstrap project 'til end of the week, I can also attach the CRD draft there :)

Flydiverny:
Could have a separate repo for just the specs as well if we want?

jonatasbaldin:
> Could have a separate repo for just the specs as well if we want?

Yeah, 💯 !

keps/20200901-standard-crd.md
```diff
@@ -0,0 +1,272 @@
...
+```yaml
+apiVersion: external-secrets.k8s.io/v1alpha1
+kind: ExternalSecret
+metadata: {...}
+
+spec:
+
+  # the amount of time before the values will be read again from the store
+  # may be set to zero to fetch and create it once
+  refreshInterval: "1h"
```
jonatasbaldin, line 122: Should we have a default value for refreshInterval? This can be a sensitive field, since it might make users reach their provider API limits.

If not, we need to mark this as required (and might give some advice on how to estimate it).

knelasevero: Default reconcile loops happen between 10h (if we don't change anything to the generated stuff), right?

keps/20200901-standard-crd.md
moolen: fixed in new repo

moolen:
I moved the CRD Spec to a dedicated repository. I fixed the `store` / `provider` naming, what is not yet addressed: 
* [1] type & runtime.RawExtension vs. typed provider https://github.com/external-secrets/kubernetes-external-secrets/pull/477#issuecomment-728827340
* [2] default value for refreshInterval https://github.com/external-secrets/kubernetes-external-secrets/pull/477#discussion_r526926784

Let's move discussion over to the new repo.

moolen:
Lemme close this here, see you on the other side :wave:

Issue #265, opened 2020-01-14
Return Error, null


debu99:
If we use the below yaml and first key "/hello-service/password1" does not exist in AWS SSM, then ExternalSecret will stuck and return Error, null

The expected action should skip the one that not exist and continue with the next one

```yaml
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  name: hello-service
spec:
  backendType: systemManager
  data:
    - key: /hello-service/password1
      name: password1
    - key: /hello-service/password2
      name: password2
```

PR #483, opened 2020-09-15
Provide a meaningful error message when an SSM parameter is missing

FaHeymann:
Related to https://github.com/godaddy/kubernetes-external-secrets/issues/265

When a requested key does not exist in AWS SSM the Externalsecret status currently reads `ERROR, null` because the AWS error has 'null' as its message. This change would add a meaningful error message.
The issue mentioned above suggests to skip the missing key. I think that would mask the underlying problem though. Let me know what you prefer.


Flydiverny (approved):
LGTM

PR #495, opened 2020-09-25
Adds azureProperty property to handle backward compatibility when using the isBinary property

### LINKED ISSUE #494: isBinary field does not work with the property field in Azure backend
According to the PR at #311, the `isBinary` field doesn't work with `property` field for Azure backend. I have a idea to solve this and I am creating this issue to track it.

SayakMukhopadhyay:
Closes #494

### DIFF
--- a/lib/backends/azure-keyvault-backend.js
+++ b/lib/backends/azure-keyvault-backend.js
@@ -36,7 +36,10 @@ class AzureKeyVaultBackend extends KVBackend {
     const secret = await client.getSecret(key)
     // Handle binary files, since the Azure client does not
     if (keyOptions && keyOptions.isBinary) {
-      return Buffer.from(secret.value, 'base64')
+      if (!keyOptions.azureProperty) {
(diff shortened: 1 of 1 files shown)


Flydiverny:
I'm not sure I understand the intent #494 references #311 which talks about `property`, this is true for all cases of returning binary data from a secret backend (some require `isBinary` to be set, some can detect automatically).

```yaml
kind: ExternalSecret
metadata:
  name: hello-service
spec:
  ...
  data:
      name: password
      property: password # <--- this is unsupported with `isBinary`
      isBinary: true
```

I'm not sure what issue or use case you are trying to cover, would be great on you could elaborate.

It's correct that the Azure backend currently doesn't have feature parity with other backends outside of `isBinary` usage, there's some info in #460 which tries to resolve this allowing `property` to be used as expected (when not using `isBinary`)

Flydiverny:
Closing this until further discussion takes place.

PR #498, opened 2020-09-26
fix: e2e tests to work with kind 0.9.0 + bump k8s version used

Flydiverny:
E2E test broke due to kind being bumped to 0.9.0 (not sure what and where the binary is provided? in the ubuntu CI image?)
api version for kind needed an update.


silasbw (approved):
:fire:

PR #548, opened 2020-11-15
feat(multitenancy): Allow to watch ExternalSecrets in specific namespaces

aabouzaid:
Hello,

This PR is 1 of 2 PRs to address support for multitenancy.

At the moment KES is not scoped, which means it works cluster-wide and there is no way to scope its access. Thus, cannot deploy 2 KES in the same cluster will different access.

So this PR allows allow to watch `externalsecrets` in specified namespaces which fix a couple of issues:
1. It allows to deploy more than 1 operator per cluster, so each tenant can deploy their own KES and optimize it according to thier workload.
2. Covers more use cases for security separation since using namespace annotation doesn't fit all use cases because in some workloads the tenant (e.g. devs) are able to create and edit namespaces.
3. Limit the access of KES service account (which used for Vault role for example) where it has gigantic access to all paths and policies in Vault for example. (this fixes issue no. #474)

I've already finished the main implementation (the last part could be supporting regex in namespaces name instead of static names), now I need to update the tests and docs.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -36,5 +36,4 @@ jobs:
           helm init --client-only
         if: matrix.helmVersion == 'V2'
       - run: ./e2e/run-e2e-suite.sh ${{ matrix.disableCustomResourceManager }} ${{ matrix.helmVersion }}
-        env:
-          DEBUG: 'true'
+
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -36,3 +36,5 @@ jobs:
           helm init --client-only
         if: matrix.helmVersion == 'V2'
       - run: ./e2e/run-e2e-suite.sh ${{ matrix.disableCustomResourceManager }} ${{ matrix.helmVersion }}
+        env:
+          DEBUG: 'true'
--- a/README.md
+++ b/README.md
@@ -262,7 +262,7 @@ metadata:
 ### Using ExternalSecret controller config
 
 ExternalSecret config allows scoping the access of kubernetes-external-secrets controller.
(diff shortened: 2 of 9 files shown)


aabouzaid:
No Idea why the E2E tests are failing :-/

Flydiverny:
Thanks for the PRs! Will probably be abit before I can take a deeper look, hopefully someone else can take a peek as well! :)
One initial concern is that `customResourceManagerDisabled` should perhaps be disabled if one watches specific namespaces to avoid having multiple deployments fighting over the CRD. 
Ideally we can also limit the RBAC permissions!

aabouzaid:
@Flydiverny That's a good point. I will add it to the readme file.

However, I think that's shouldn't be done automatically (i.e. when namespaces are watched, automatically disable CRD management) but let the user set it because I could want to make KES working with a set of namespaces but it's still the only KES in the cluster.

aabouzaid:
@Flydiverny I've updated the branch to fix the conflict after latest changes (about timeout) :-)

jonathonbattista:
Any update on this

Flydiverny:
Love to get this merged but with E2E tests failing its blocked.
Will try to look at it when I can, but if someone else figures it out or can look into it please do help to get this merged!

ginoh:
🎉

facundo-tul:
i have a vault role binded to a service account and a namespace with specific access to secrets.
running KES i need to give it access with a specific role asigned to it.
when i try to generate a new external-secret with a specific role on another namespace i have `namespace not authorized` error.
the only way to make it work is give KES unrestricted accest to all my secrets to put it on multiple namespaces.
the idea to have multiple instances running implies houndred of kes pod doing nothing.
what i am missing wich is the correct way to handle this?

PR #559, opened 2020-11-17
fix: annotation provided as list is not supported by k8s

nbendafi-yseop:
As reported by @ronlut (cf. https://github.com/external-secrets/kubernetes-external-secrets/pull/409#issuecomment-728855887), annotations value _should_ be of string type.

This PR and use of Yaml mutli-line string for `externalsecrets.kubernetes-client.io/permitted-key-name`
```bash
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Namespace
metadata:
  name: core-namespace
  annotations:
    # annotation key is configurable
    externalsecrets.kubernetes-client.io/permitted-key-name: |
      - "/dev/cluster1/core-namespace/.*"
      - "/common/.*"
EOF
```
should work.

### DIFF
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -2,6 +2,7 @@
 
 const clonedeep = require('lodash.clonedeep')
 const merge = require('lodash.merge')
+const yaml = require("js-yaml")
 
 /**
  * Kubernetes secret descriptor.
@@ -211,8 +212,13 @@ class Poller {
     // 1. testing naming convention if defined in namespace
 
     const externalData = descriptor.data || descriptor.properties
-    const namingConvention = namespace.metadata.annotations[this._namingPermittedAnnotation]
+    let namingConvention = namespace.metadata.annotations[this._namingPermittedAnnotation]
     let reNaming = new RegExp()
+    if (!Array.isArray(namingConvention)) {
+      // Unmarshal YAML
+      namingConvention = yaml.safeLoad(namingConvention)
+    }
+
     if (Array.isArray(namingConvention)) {
       reNaming = new RegExp(namingConvention.join('|'))
     } else {
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -217,6 +217,13 @@ class Poller {
     if (!Array.isArray(namingConvention)) {
       // Unmarshal YAML
       namingConvention = yaml.safeLoad(namingConvention)
+      if (!Array.isArray(namingConvention) && typeof namingConvention !== 'string') {
+        allowed = false
+        reason = `Fail to parse ${this._namingPermittedAnnotation} on namespace ${namespace.metadata.name}.`
+        return {
+          allowed, reason
+        }
+      }
     }
 
     if (Array.isArray(namingConvention)) {
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -2,7 +2,7 @@
 
 const clonedeep = require('lodash.clonedeep')
 const merge = require('lodash.merge')
-const yaml = require("js-yaml")
+const yaml = require('js-yaml')
 
 /**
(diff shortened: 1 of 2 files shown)


Flydiverny (commented):
Code seems fine. Not fully sure how I feel about the usage, eg annotation with a yaml formatted value.

We should probably throw a fit if the parsed value is not on the expected format? 
Eg if namingConvention doesnt result in an array or a plain string after the yaml parse. Not sure what happens if one where to provide an object.

nbendafi-yseop:
@Flydiverny Would something like
```diff
diff --git a/lib/poller.js b/lib/poller.js
index 38c0736..51535c9 100644
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -217,6 +217,13 @@ class Poller {
     if (!Array.isArray(namingConvention)) {
       // Unmarshal YAML
       namingConvention = yaml.safeLoad(namingConvention)
+      if (!Array.isArray(namingConvention) && typeof namingConvention != 'string') {
+        allowed = false
+        reason = `Fail to parse ${this._namingPermittedAnnotation} on namespace ${namespace.metadata.name}.`
+        return {
+          allowed, reason
+        }
+      }
     }
 
     if (Array.isArray(namingConvention)) {
```
be enough as a safety belt ?

Flydiverny:
> @Flydiverny Would something like
> 
> ```diff
> diff --git a/lib/poller.js b/lib/poller.js
> index 38c0736..51535c9 100644
> --- a/lib/poller.js
> +++ b/lib/poller.js
> @@ -217,6 +217,13 @@ class Poller {
>      if (!Array.isArray(namingConvention)) {
>        // Unmarshal YAML
>        namingConvention = yaml.safeLoad(namingConvention)
> +      if (!Array.isArray(namingConvention) && typeof namingConvention != 'string') {
> +        allowed = false
> +        reason = `Fail to parse ${this._namingPermittedAnnotation} on namespace ${namespace.metadata.name}.`
> +        return {
> +          allowed, reason
> +        }
> +      }
>      }
>  
>      if (Array.isArray(namingConvention)) {
> ```
> 
> be enough as a safety belt ?

Yepp! Seems good!
Nit: switch `!=` to `!==` :)

ErikLundJensen:
Why not just use the regex for giving multiple matches like this:
```
externalsecrets.kubernetes-client.io/permitted-key-name:
  "/dev/cluster1/core-namespace/.*|/common/.*"
  ```
  
  That works without any changes to the code base.
  I suggest that this pull-request is rejected.

nbendafi-yseop:
> Why not just use the regex for giving multiple matches like this:

Yes, that's what documentation says (https://github.com/external-secrets/kubernetes-external-secrets#scoping-access). The goal of this PR, is to have a more _user-friendly_ (more readable ?) ways to configure permitted key names. Though, _only_ cosmetic.

ErikLundJensen:
But the proposed change does not make it more user friendly. Having a list in an annotation is very unnormal. It might give problems in other tools that process annotation (example logging). At the same time it complicates the code base with without any real benefit.

I wote for closing this PR as it breaks the idea of annotations are strings, just like labels. Furthermore it breaks the YAML structure as the new list is not a real yaml list but just something that looks like yaml.

nbendafi-yseop:
@ErikLundJensen 
> I wote for closing this PR as it breaks the idea of annotations are strings, just like labels. Furthermore it breaks the YAML structure as the new list is not a real yaml list but just something that looks like yaml.

That's enough argument to convince me. Closing PR.

Flydiverny:
Aight! Thanks for feedback @ErikLundJensen and sorry for leaving this hanging for so long 😄

PR #567, opened 2020-11-25
feat: enable proxy configuration

derdanne:
We use AWS EKS behind HTTP Proxy and use the IRSA feature with AWS Secrets Manager. Since there is no OIDC endpoint available as VPC endpoint, we need to pass a proxy configuration to get a valid session token. The 'global-agent' module provides an easy way to implement HTTP_PROXY as well as NO_PROXY configuration to the application.

* enable proxy configuration via environment variables with global-agent
* enable proxy configuration in Dockerfile

Signed-off-by: Daniel Klockenkämper <daniel.klockenkaemper@vodafone.com>


Flydiverny:
Please reopen / update if not solved by the proxy feature merged in #601

PR #585, opened 2020-12-22
feat: add general support for isBinary for all backends

ninadpage:
## Changes
- Add support for `isBinary` in `KVBackend` + tests
- Remove specific implementation of `isBinary` from Azure Key Vault & GCP Secrets Manager backends
- Update description for `isBinary` field in the CRD to remove Azure-specific details
- Update docs

## Background
- Support for binary secrets was missing in Hashicorp Vault backend
- #311 (Azure) and #353 (GCP) added support for `isBinary` for specific backends, both using the same approach
- However, that could not be replicated for Hashicorp Vault because there is no concept of "text strings" secrets in Vault - all secrets hold key-value pairs. Since `property` is not available in `VaultBackend`, you cannot handle binary secrets there either.
- After fixing #460 (Azure) and #373 (GCP), all backends return a response of same type(s) and structure. So it is now possible to write a general implementation for `isBinary` in `KVBackend`.
- This also removes the restriction that you cannot combine `isBinary` with `property` - it is possible (and in fact a requirement for Hashicorp Vault).

### DIFF
--- a/lib/backends/kv-backend.test.js
+++ b/lib/backends/kv-backend.test.js
@@ -490,6 +490,7 @@ describe('kv-backend', () => {
       kvBackend._get.onCall(0).resolves(Buffer.from('YmluYXJ5Cg==', 'base64')) // base64 of binary as a Buffer
       kvBackend._get.onCall(1).resolves(Buffer.from('YmluYXJ5Cg==', 'base64')) // base64 of binary as a Buffer
       kvBackend._get.onCall(2).resolves('YmluYXJ5Cg==') // base64 of binary as String
+      kvBackend._get.onCall(3).resolves('test')
       // e.g. AWS Secrets Manager will return `SecretBinary` as a Buffer and `SecretString` as a String
 
       const manifestData = await kvBackend.getSecretManifestData({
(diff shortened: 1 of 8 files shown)


lib/backends/kv-backend.test.js
```diff
@@ -446,5 +485,38 @@ describe('kv-backend', () => {
...
+          }]
+        }
+      })
+
+      expect(manifestData).deep.equals({
+        binaryPropertyName1: 'YmluYXJ5Cg==', // base64 of binary (unchanged)
+        binaryPropertyName2: 'YmluYXJ5Cg==', // base64 of binary (unchanged)
+        stringPropertyName3: 'WW1sdVlYSjVDZz09' // base64 of base64 of binary
+      })
+    })
```
Flydiverny, line 520: Can we add a test variant which explicitly sets `isBinary: false` where the value is a string?

Flydiverny (commented):
Looks great! Thanks for this contribution! 👍 
I made a small comment. Since there's some logic on determining if `isBinary` is true or false, it would be nice to have a test which explicitly sets it as false.

Flydiverny:
Looks like e2e tests are failing due to using old repo URLs I'll see if I can amend that.

Flydiverny (approved):

lib/backends/kv-backend.test.js
ninadpage: @Flydiverny Makes sense, done!

PR #592, opened 2021-01-05
feat: adds lodash get function for nested keys lookup

jonathonbattista:
Currently, with the Vault backend, only top-level properties can be used for formatting Secrets: [https://github.com/external-secrets/kubernetes-external-secrets/blob/master/lib/backends/kv-backend.js#L47](https://github.com/external-secrets/kubernetes-external-secrets/blob/master/lib/backends/kv-backend.js#L47)

We would like to be able to support nested property lookups for more complex Vault JSON secrets.

I added a simple lodash implementation to check whether the nested property exists, then fetch the value if it does.

### DIFF
--- a/lib/backends/kv-backend.js
+++ b/lib/backends/kv-backend.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const AbstractBackend = require('./abstract-backend')
-const get = require('lodash/get')
+const { get, hasIn } = require('lodash')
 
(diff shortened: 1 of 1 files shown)


Flydiverny (commented):
Can we add a note to the readme and some examples to highlight the existence of this feature?
Would also need to add a test case that tries to fetch a deeply nested key.

PR #601, opened 2021-01-15
Pass proxy to AWS SDK

markthebault:
Hi All,

I aws trying to debug why I could not connect to AWS endpoints with an environment that does not use the Internet (private environment). I figured out that proxy settings need to be passed explicitly to AWS SDK: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/node-configuring-proxies.html

I am not a NodeJS developer, so feel free to help me out if you think this code is not good enough.

Best,
Mark.


Flydiverny (approved):
Thanks!

PR #602, opened 2021-01-16
feat(aws): allow custom endpoints for aws services

smcavallo:
Allow custom endpoints for aws services

Useful if you need to use the FIPS endpoints ->  https://docs.aws.amazon.com/general/latest/gr/asm.html
Or less likely use case if you need to override to use VPC endpoints when using custom DNS -> https://docs.aws.amazon.com/secretsmanager/latest/userguide/vpc-endpoint-overview.html

kubernetes-external-secrets is amazing and FIPS support is required for running it in fedRAMP compliant environments.

Signed-off-by: S.Cavallo <smcavallo@hotmail.com>

### DIFF
--- a/config/aws-config.js
+++ b/config/aws-config.js
@@ -18,11 +18,27 @@ const localstack = process.env.LOCALSTACK || 0
 
 const intermediateRole = process.env.AWS_INTERMEDIATE_ROLE_ARN || 0
 
+const sts_endpoint = process.env.AWS_STS_ENDPOINT || 0
+const ssm_endpoint = process.env.AWS_SSM_ENDPOINT || 0
+const sm_endpoint = process.env.AWS_SM_ENDPOINT || 0
+
 let secretsManagerConfig = {}
 let systemManagerConfig = {}
 let stsConfig = {
   region: process.env.AWS_REGION || 'us-west-2',
   stsRegionalEndpoints: process.env.AWS_STS_ENDPOINT_TYPE || 'regional'
+}
+
+if (sm_endpoint) {
+  secretsManagerConfig.endpoint = sm_endpoint
+}
+
+if (ssm_endpoint) {
+  systemManagerConfig.endpoint = ssm_endpoint
+}
+
+if (sts_endpoint) {
+  stsConfig.endpoint = sts_endpoint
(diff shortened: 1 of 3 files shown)


config/aws-config.js
```diff
@@ -18,13 +18,29 @@ const localstack = process.env.LOCALSTACK || 0
...
+
+if (ssmEndpoint) {
+  systemManagerConfig.endpoint = ssmEndpoint
+}
+
+if (stsEndpoint) {
+  stsConfig.endpoint = stsEndpoint
+}
+
 if (localstack) {
```
moolen, line 44: I'd like to consolidate the configuration for e2e tests and your custom endpoints using `AWS_*_ENDPOINT`. This way we can hit two birds with one stone:
1. have end-to-end tests for your feature
2. simplify the configuration code

I think it makes sense to use your `AWS_*_ENDPOINT` in the [e2e test code](https://github.com/external-secrets/kubernetes-external-secrets/search?q=LOCALSTACK_SM_URL) and remove the localstack/e2e-specific confinguration variables.

What do you think?

moolen (changes requested):
Hey, thanks for bringing this up, i totally support your idea! I have one suggestion about testing your feature. Please let me know what you think.

config/aws-config.js
smcavallo: @moolen - that totally makes sense.  Is it possible to merge this PR and then consolidate the e2e test endpoints in a separate PR?  This PR makes the endpoints configurable and the next PR can change the e2e tests to leverage this configuration.
It will be cleaner this way as there will be some questions as to how and where to best make the e2e changes, which would be easier to review as part of a separate change.

moolen: > consolidate the e2e test endpoints in a separate PR?

Yes, for sure! 
I just noticed the `AWS_*_ENDPOINT` aren't yet added to the helm chart. Could you please add/document them there aswell?

smcavallo: I added the other env vars to the readme and commented in the default values.yaml

moolen (approved):
LGTM! Nice, thank you very much for your contribution!

PR #610, opened 2021-01-28
add stale workflow to auto-close issues

moolen:
This PR adds a workflow which runs daily to:
* mark issues without activity as `stale` after 90d
* close stale issues after 30d with no activity

See [actions/stale](https://github.com/actions/stale).

@Flydiverny could you plase take a look at this :heart:  ?

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,17 @@
+name: 'Close stale issues and PR'
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


Flydiverny (approved):
LGTM 👍 
Will this go nuts on all existing things? Maybe that's fine tho as its the only way to get rid of old things 😄 Guess I should prepare for lots of notifications lol.

PR #615, opened 2021-02-02
feat(helm): add in ability to inject init containers in to deployment from values

chrismellard:


charts/kubernetes-external-secrets/values.yaml
```diff
@@ -123,3 +123,9 @@ serviceMonitor:
   enabled: false
   interval: "30s"
   namespace:
+
+deploymentInitContainers: {}
+
+emptyDirVolumes: {}
```
Flydiverny, line 129: Can we do more generic and add `extraVolumeMounts` and `extraVolumes` instead ? :)

chrismellard: Updated the PR to reflect your suggestion. Thanks for the review 👍

charts/kubernetes-external-secrets/values.yaml
```diff
@@ -125,3 +125,19 @@ serviceMonitor:
...
+# Add in additional named volumes and volume mounts to the deployment
+#
+#extraVolumes:
+#  namedVolume:
+#    emptyDir: {}
+#
+#extraVolumeMounts:
+#  namedVolume:
+#    mountPath: /usr/path
+#    readOnly: false
```
Flydiverny, line 140:
```suggestion
extraVolumes: []
#  - name: namedVolume
#    emptyDir: {}
#
extraVolumeMounts: []
# - name: namedVolume
#    mountPath: /usr/path
#    readOnly: false
```

Flydiverny (commented):
Think we can keep it even more "standard". At least I think this pattern is pretty common in charts. I could be wrong a bit out of the loop 😄

charts/kubernetes-external-secrets/values.yaml
chrismellard: All good - more likely me out of the loop ;) Updated.

Flydiverny (approved):
👍 merging assuming e2e tests doesnt magically fail

chrismellard:
Hold off. spotted one problem

chrismellard:
Resolved. Was outputting volumes yaml in volumeMounts block :)

PR #641, opened 2021-02-25
Enhance info log message about region for AWS SM backend

cebidhem:
This pull requests adds the following enhancement:

- An accurate info log message when the AWS region is not specified in the ExternalSecret manifest but uses the `AWS_REGION` env var as explained in https://github.com/external-secrets/kubernetes-external-secrets/issues/635

The idea is not to mislead the users by throwing an undefined in an info log message.

It is very small I know, but this simple addition can help not having any doubt about if external-secret is doing properly the task or not.


Flydiverny (approved):
Thanks for taking the time! :)

Issue #630, opened 2021-02-21
Retrieving Secrets from Vault throws 404 without any error text.


abhinavkhanna-sf:
Retrieving Secrets from Vault throws 404 without any error text.

`"payload":{"response":{"statusCode":404,"body":{"errors":[]}}},"msg":"failure while polling the secret <ns>/<external-secret>`

I am unable to debug what I am doing wrong. 

edit:

more logs

```
{"level":30,"time":1613896610435,"pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5f45dd675d-lz9jf","msg":"running poll on the secret it-automation/gh-secret"}
{"level":20,"time":1613896610447,"pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5f45dd675d-lz9jf","msg":"fetching new token from vault for role it-automation-runner-role on it/kubernetes"}
{"level":20,"time":1613896610450,"pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5f45dd675d-lz9jf","msg":"fetching new token from vault for role it-automation-runner-role on it/kubernetes"}
{"level":20,"time":1613896610450,"pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5f45dd675d-lz9jf","msg":"fetching new token from vault for role it-automation-runner-role on it/kubernetes"}
{"level":50,"time":1613896610538,"pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5f45dd675d-lz9jf","payload":{"response":{"statusCode":404,"body":{"errors":[]}}},"msg":"failure while polling the secret it-automation/gh-secret"}
```

sklarsa:
I'm experiencing the same behavior, but this is when trying to sync kv v2 secrets

Flydiverny:
`"statusCode":404` Seems like it can't find the given secret? 😄

sklarsa:
Sorry @Flydiverny, still adding context, but `vault kv get secret/data/hello-service/credentials` works when running on a separate pod with the same serviceaccount

Flydiverny:
Maybe its failing to get the token, judging by that being the last message before the errors. In that case maybe the vault mount point? But I honestly don't know to much about vault regardless 😄

sklarsa:
No prob! I'm still doing what I can to debug.  The auth seems to be working with `kv v1` secrets, I'm just running into a different error (wanted to keep this ticket clean).
 (With v1 secrets and same yaml)
[...]

PR #642, opened 2021-02-25
fix(vault): override api version for kv v2

Flydiverny:
Not sure if this fixes things or not but was looking at #630 and https://github.com/kr1sp1n/node-vault/issues/82

Test image:
docker.io/flydiverny/kubernetes-external-secrets:vault-kv2-test

(this is untested, test at your own leisure)

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -77,7 +77,7 @@ if (envConfig.vaultNamespace) {
     'X-VAULT-NAMESPACE': envConfig.vaultNamespace
   }
 }
-const vaultFactory = () => vault(vaultOptions)
+const vaultFactory = (overrides = {}) => vault({ ...vaultOptions, ...overrides })
 
 // The Vault token is renewed only during polling, not asynchronously. The default tokenRenewThreshold
 // is three times larger than the pollerInterval so that the token is renewed before it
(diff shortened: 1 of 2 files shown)


sklarsa:
Does this change the global vault configuration for all vault-backed secrets? Because it is possible to use both v1 and v2 of the secrets engines at different mount points

sklarsa:
I don't think this is doing the trick...  I cherry-picked your commit into 6.1.0 (because I'm having trouble running master) and it looks like this particular 404 is stemming from token auth
```
{"level":20,"time":1614285380849,"pid":927,"hostname":"k3d-app-server-0","msg":"fetching new token from vault for role demo on kubernetes"}
{"level":20,"time":1614285380857,"pid":927,"hostname":"k3d-app-server-0","msg":"fetching new token from vault for role demo on kubernetes"}
{"level":50,"time":1614285380876,"pid":927,"hostname":"k3d-app-server-0","payload":{"response":{"statusCode":404,"body":{"errors":[]}}},"msg":"failure while polling the secret default/hello-vault-service"}
```
Changing the secret engine type shouldn't impact the token auth, so I'll look into this some more

Issue #643, opened 2021-02-25
Vault kv v1 issue: "Failed to JSON.parse value for 'secret/path', please verify that your secret value is correctly formatted as JSON..."


sklarsa:
As referenced in [another ticket](https://github.com/external-secrets/kubernetes-external-secrets/issues/630#issuecomment-786051315), I'm hitting an issue when using vault kv v1:

`Failed to JSON.parse value for 'kv/data/hello-service/credentials', please verify that your secret value is correctly formatted as JSON. To use plain text secret remove the 'property: password'`

Environment
-------------
kubernetes-external-secrets v6.1.0
vault 1.6.2

external-secret.yaml
```
apiVersion: 'kubernetes-client.io/v1'
kind: ExternalSecret
metadata:
  name: hello-vault-service
spec:
  backendType: vault
  vaultMountPoint: kubernetes
  vaultRole: demo
  data:
  - name: passwordv1
    key: kv/data/hello-service/credentials
    property: password
  - name: api-keyv1
    key: kv/data/hello-service/credentials
    property: api-key
```

Detail
------

This seems like a problem that the application is not properly recognizing vault kv v1 secrets, and is treating them as v2.  I added the following debugging line to `vault-backend.js` under `const secretResponse = await client.read(key)`
```
this._logger.info(`kvVersion ${kvVersion}`)
```
and received the following output:
```
{"level":30,"time":1614279814580,"pid":435,"hostname":"k3d-app-server-0","msg":"kvVersion 2"}
```

The secret does get created, it's just empty
```
$ k describe secret hello-vault-service
Name:         hello-vault-service
Namespace:    default
Labels:       <none>
Annotations:  <none>

Type:  Opaque

Data
====
```

sklarsa:
This is because kv version defaults to 2
https://github.com/external-secrets/kubernetes-external-secrets/blob/939431695d7e4fd1b2fe2422692773b7a076afd9/lib/backends/vault-backend.js#L44

Based on the fact that the [node-vault](https://github.com/kr1sp1n/node-vault/issues/82) dependency doesn't support kv2, I don't think this package can support it, and think that this should default back to v1 until either the upstream package gets patched or `kubernetes-external-secrets` makes modifications (or an entirely new) vault client.

SharifiFaranak:
I ran into this issue when trying to use k8s external secrets to fetch a password from the dynamic secret engine.
[...]

PR #644, opened 2021-02-25
Fixes Vault kv v1 issue #643 

sklarsa:
Sets the default vault kv engine to version 1 instead of 2.  Until [kv versioning is fixed](https://github.com/kr1sp1n/node-vault/issues/82) in the upstream node-vault package, this should default to the working version.

closes https://github.com/external-secrets/kubernetes-external-secrets/issues/643

### DIFF
--- a/lib/backends/vault-backend.js
+++ b/lib/backends/vault-backend.js
@@ -11,7 +11,6 @@ class VaultBackend extends KVBackend {
    * @param {Object} logger - Logger for logging stuff.
    */
   constructor ({ vaultFactory, tokenRenewThreshold, logger, defaultVaultMountPoint, defaultVaultRole }) {
-    debugger
     super({ logger })
     this._vaultFactory = vaultFactory
     this._clients = new Map()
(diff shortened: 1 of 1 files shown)


Flydiverny:
As this would be a breaking change I'm not sure if we should apply it. I agree with the motivation that it makes more sense to default to the working version.
But you can also easily set the `kvVersion` in your ExternalSecret

```yaml
kind: ExternalSecret
metadata:
  name: hello-vault-service
spec:
  backendType: vault
  vaultMountPoint: kubernetes
  vaultRole: demo
  kvVersion: 1
  data:
  - name: passwordv1
    key: kv/data/hello-service/credentials
    property: password
  - name: api-keyv1
    key: kv/data/hello-service/credentials
    property: api-key
```

SharifiFaranak:
I ran into this issue when trying to use k8s external secrets to fetch a password from the dynamic secret engine.
`"msg":"Failed to JSON.parse value for 'test-postgres/static-creds/test-static-role', please verify that your secret value is correctly formatted as JSON. To use plain text secret remove the 'property: password'"}` 
Is there a way to also specify the secret engine type in the `ExternalSecret` definition like how we can set `kvVersion: 1`?

[The docs](https://github.com/external-secrets/kubernetes-external-secrets#hashicorp-vault) don't really mention that.
Thanks!!

PR #648, opened 2021-03-04
Enhance info log message about region for AWS System Manager backend

yutachaos:
## Description
- The secret manager was supported in the PR below, but it was supported because the system manager is not supported.

## FYI 
- https://github.com/external-secrets/kubernetes-external-secrets/pull/641


moolen (approved):
It makes sense to align the log messages :) 
Thank you for fixing that ❤️!

PR #651, opened 2021-03-08
Update base image from 12.19.0-alpine to 12.21.0-alpine

Nick-Triller:
Hi,

this PR updates the Docker base image to the latest node 12 LTS minor release available on Docker Hub.

12.x is in maintenance mode until 2022-04-30. However, it should be considered to switch to the active LTS release 14.x which will be in maintenance mode until 2023-04-30. I am not sure how many breaking changes were introduced by node between 12.x and 14.x.

Best
Nick

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.19.0-alpine
+FROM node:12.21.0-alpine
 
 ENV NODE_ENV production
 ENV NPM_CONFIG_LOGLEVEL info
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:12.19.0-alpine
+FROM node:12.21.0-alpine
```
moolen, line 1: Hey Nick thanks for bringing this up! I think we can straight-up use `12-alpine`. That way we don't need to care about minor updates in the future.

```suggestion
FROM node:12-alpine
```
I'm also open to update node to 14, but that might be a little bit of a bigger task: checking deps, checking used functions in different secret store implementations etc..

moolen (approved):
thank you :)

PR #655, opened 2021-03-12
feat: Check if the secrets has changed before reading it when setting parameter store as backendtype

martingaleano:
https://github.com/external-secrets/kubernetes-external-secrets/issues/654

This pull request add the next feature:

When we set the backendType as parameterStore, the controller check the history of the parameter (if the environment variable SSM_CHECK_UPDATED is set):

    if the value has changed before the interval (POLLER_INTERVAL_MILLISECONDS), the controller update the secret
    if the value has changed after the interval (POLLER_INTERVAL_MILLISECONDS), the controller does not update the secret

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -32,7 +32,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Setup node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2.1.5
         with:
           node-version: 12
       - run: npm install


martingaleano:
This pull request add the next feature:

When we set the backendType as parameterStore, the controller check the history of the parameter (if the environment variable  SSM_CHECK_UPDATED is set):

- if the value has changed before the interval (POLLER_INTERVAL_MILLISECONDS), the controller update the secret
- if the value has changed after the interval (POLLER_INTERVAL_MILLISECONDS), the controller does not update the secret

lib/backends/system-manager-backend.js
```diff
@@ -49,13 +51,24 @@ class SystemManagerBackend extends KVBackend {
       client = this._clientFactory(factoryArgs)
     }
     try {
-      const data = await client
-        .getParameter({
+      const check = await client
+        .getParameterHistory({
```
moolen, line 55: `getParameterHistory()` does not return all results, it returns at max 50. Using `NextToken` you can fetch more.

moolen (commented):
I think it makes sense to factor out a, let's call it `"parameter-store-cache"`, that tracks and refreshes the history of all parameters for the scope of each poller (we can't do it globally because every poller instance can have it's own set of credentials/roles).

Is `getParameterHistory()` a free API call? I think right now it will make an API call with every refresh (fetch history) and an additional one to fetch the parameter value. 

I would really appreciate it if you could add e2e tests for this feature, maybe even add a load-test to check that we support  more than ~1k (or so?) parameters because we store them in-memory.

lib/backends/system-manager-backend.js
martingaleano: Thanks for your answer, i'm really excited of contributing to the project.

The getParameterHistory() is a free API call if you use Standard parameter. (https://aws.amazon.com/systems-manager/pricing/#Parameter_Store)

I'm adding e2t tests and using NextToken option. I will push the changes soon. Thanks for your review.

martingaleano: Hi @moolen i've updated the PR adding e2e test (with 500 parameters because when i tested with 1K it gives me TimeOutError).

I also added NextToken to the request so i can fetch more secret.

Thanks

e2e/tests/ssm.test.js
```diff
@@ -216,5 +217,95 @@ describe('ssm', async () => {
       expect(result).to.not.equal(undefined)
       expect(result.body.status.status).to.contain('namespace does not allow to assume role let-me-be-root')
     })
+    it('should not pull from ssm after waiting POLLER_INTERVAL_MILLISECONDS', async () => {
+      backends.systemManager.checkUpdated = true
```
moolen, line 221: The controller runs in a different pod/process than the e2e test. You can't change the behavior of it like this.

e2e/tests/ssm.test.js
```diff
@@ -216,5 +217,95 @@ describe('ssm', async () => {
...
+                {
+                  key: `/e2e/ssmcheck/${uuid}`,
+                  name: 'name'
+                }
+              ]
+            }
+          }
+        })
+
+      await new Promise(resolve => setTimeout(resolve, 5000))
```
moolen, line 255: This shouldn't be necessary, i think.

e2e/tests/ssm.test.js
```diff
@@ -216,5 +217,95 @@ describe('ssm', async () => {
...
+        .get()
+      expect(result).to.not.equal(undefined)
+      expect(result.body.status.status).to.contain('The parameter has not changed since last time')
+    })
+    it('should pull from ssm with 500 history of parameters', async () => {
+      backends.systemManager.checkUpdated = true
+
+      for (let i = 1; i <= 500; i++) {
+        await putParameter({
+          Name: `/e2e/ssmcheck/${uuid}`,
```
moolen, line 273: To properly test the parameter history pagination and in-memory caching behavior with many many parameters you would need to variablize the name to, e.g. `/e2e/ssmcheck/${uuid}/${i}` instead of overriding the value.

lib/backends/system-manager-backend.js
```diff
@@ -49,13 +51,34 @@ class SystemManagerBackend extends KVBackend {
...
-      const data = await client
-        .getParameter({
-          Name: key,
-          WithDecryption: true
-        })
+      const params = {
+        Name: key,
+        WithDecryption: false
+      }
+      const check = await client
```
moolen, line 58: You assign `check` once but subsequent calls with `NextToken` return an array of parameter metadata that must be merged

moolen (changes requested):
Hey @martingaleano thank you for your work!
I think it is needed to factor out a Service that stores metadata per account & role. Otherwise we'll do more API calls than less.

lib/backends/system-manager-backend.js
```diff
@@ -49,13 +51,34 @@ class SystemManagerBackend extends KVBackend {
...
         .promise()
-      return data.Parameter.Value
+
+      while (check.NextToken) {
+        params.NextToken = check.NextToken
+        await client.getParameterHistory(params).promise()
+      }
+      const timetocompare = new Date(Date.now() - this._intervalPolling)
+
+      if (check.Parameters[check.Parameters.length - 1].LastModifiedDate > timetocompare || !this._checkUpdated) {
```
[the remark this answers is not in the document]
martingaleano: Hi @moolen. I think your idea is better, thanks for the suggestion.

I did this draw and please tell me if i got your idea:
image

1- The ParameterMetadaService is an infinite loop that store (in memory?) the parameter history
2- The SystemManagerBackend consume the API explained in 1.
3. If the parameter has changed, SystemManagerBackend fetch the new value

Is there any similar code in this project that i can be based on?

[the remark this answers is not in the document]
moolen: Hey! yeah, that's pretty much what i've thought of! Yes, it probably needs to be in-memory (hence the load test). 
> Is there any similar code in this project that i can be based on?

There's nothing i'm aware of. I imagine a

```javascript
cass PMS {
  constructor(interval){
    this._interval = interval;
    this.data = {};
  }

  // this does the hard work
  fetchHistory(){
    /*
    for every N seconds; do:
      for each account/region combination that was previously seen; do:
        1. fetch history of all parameters
        2. store them in this.data 
    */
  }
  
  // this is the cheap lookup
  hasChanged(account, region, param, since){
    // the lookup key is important, the controller can handle multiple accounts/roles
    if data[`${account}/${role}`][param].LastModifiedDate > since {
      return true
    }
    return false
  }
}
```

PR #656, opened 2021-03-14
ibm cloud support

No9:
Hi All
Thanks for providing a such a well laid out project it's been a pleasure to work on.
This PR adds support to integrate the secrets manager service on IBM Cloud with Kubernetes.
It largely consists of a new backend that follows the AliCloud environment variables pattern along with the relevant enum updates.
As well as the backend mock test I have also tested it on an IKS instance and everything works as expected.
Hopefully I haven't missed anything too obvious and I'd appreciate any feedback.


Flydiverny:
Cool! I'll try to find some time to look thru this :)

package.json
```diff
@@ -35,6 +35,7 @@
     "@azure/identity": "^1.0.3",
     "@azure/keyvault-secrets": "^4.0.4",
     "@google-cloud/secret-manager": "^3.2.3",
+    "@ibm-cloud/secrets-manager": "^0.0.6",
```
Flydiverny, line 38: Looks like a newer version 0.1.0 is no available, might be worth bumping? 😄

Flydiverny (commented):
While I have no way to test this, I dont see anything strange :) 

package-lock has got a conflict due to another PR merge. could you rebase and possibly bump the added dependency or was there a reason to not using the latest version? (its pretty recent so maybe you did this before it was available 😄)

package.json
No9: I'll take a look at bumping this and test it.

No9: So it looks like 0.1.0 release hasn't made it into npm yet so this can't land yet. 
We should get it in so I'll follow up with @IdanAdar to see when it will land in the registry.

IdanAdar: Published.

No9: Hey @Flydiverny Patch applied but the container has CVEs do I need to fix this?

Flydiverny: @No9 thanks for the update, no need to address the CVE. I was looking at it earlier and there's a fix out already. But it seems our pipeline is pulling an old base image.

Flydiverny (approved):

No9:
Thanks for the merge and pushing this through @Flydiverny I
FYI  had to bump the image version to master in the values.yml file as the versioned image isn't released yet but everything works as expected.

Flydiverny:
Yepp! Was considering to add some more changes to next release but I'll probably try to make one today or tomorrow.

Issue #401, opened 2020-06-05
Log lines might end up with duplicate JSON keys


bgdnlp:
In some situations the log JSON will have duplicate keys, I assume because the "message" dict/map is merged with the log one. For example, the "time" key is included twice here:
```
{
  "level":50,
  "time":1591347107037,
  "pid":19,
  "hostname":"kubernetes-external-secrets-6799c84569-kfjcj",
  "message":"Access to KMS is not allowed",
  "code":"AccessDeniedException",
  "time":"2020-06-05T08:51:47.036Z",
  "requestId":"1894b6c9-b698-4657-a3cb-ecd8f044afee",
  "statusCode":400,
  "retryable":false,
  "retryDelay":77.07928574264942,
  "stack":"AccessDeniedException: Access to KMS is not allowed\n    at Request.extractError (/app/node_modules/aws-sdk/lib/protocol/json.js:51:27)\n    at Request.callListeners (/app/node_modules/aws-sdk/lib/sequential_executor.js:106:20)\n    at Request.emit (/app/node_modules/aws-sdk/lib/sequential_executor.js:78:10)\n    at Request.emit (/app/node_modules/aws-sdk/lib/request.js:683:14)\n    at Request.transition (/app/node_modules/aws-sdk/lib/request.js:22:10)\n    at AcceptorStateMachine.runTo (/app/node_modules/aws-sdk/lib/state_machine.js:14:12)\n    at /app/node_modules/aws-sdk/lib/state_machine.js:26:10\n    at Request.<anonymous> (/app/node_modules/aws-sdk/lib/request.js:38:9)\n    at Request.<anonymous> (/app/node_modules/aws-sdk/lib/request.js:685:12)\n    at Request.callListeners (/app/node_modules/aws-sdk/lib/sequential_executor.js:116:18)",
  "type":"Error",
  "msg":"failure while polling the secret kubernetes-external-secrets/test"
}
```

pierluigilenoci:
This problem happens because Docker itself add a "time" field inside its log when merged. If you want to produce a JSON log you need to use a different "time" field name. I will suggest avoiding any docker/logstash/fluentbit/fluentd/elasticsearch "reserved" key (example: "@timestamp")

pierluigilenoci:
@silasbw @Flydiverny can you please take a look?

pierluigilenoci:
@Niksko @moolen @megakid can you help in some way?

moolen:
[...]

PR #660, opened 2021-03-17
fix!: rename time field to avoid duplicate time key in log output

pierluigilenoci:
Ref: #401

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -42,7 +42,8 @@ const logger = pino({
       return { level: envConfig.useHumanReadableLogLevels ? label : number }
     }
   },
-  nestedKey: 'payload'
+  nestedKey: 'payload',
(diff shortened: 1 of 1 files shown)


config/index.js
```diff
@@ -42,7 +42,8 @@ const logger = pino({
       return { level: envConfig.useHumanReadableLogLevels ? label : number }
     }
   },
-  nestedKey: 'payload'
+  nestedKey: 'payload',
+  timestamp: () => `,"message_time":"${new Date(Date.now()).toISOString()}"`
```
Flydiverny, line 46: For reference: https://github.com/pinojs/pino/blob/master/docs/api.md#timestamp-boolean--function

Flydiverny (approved):
LGTM! Was considering whether or not we should just have a config to turn it on or off, or I guess rename it and keep the current time key as default.
But I think we can just go with this.

PR #663, opened 2021-03-18
feat!: drop helm v2 and builtin CRD management

Flydiverny:
Since Helm 2 is EOL this drops helm 2 support and simplifies docs and removes the built in CRD handling depending entirely on them being installed by the user manually or thru helm.


bin/daemon.js
```diff
@@ -34,7 +33,8 @@ async function main () {
   logger.info('loading kube specs')
   await kubeClient.loadSpec()
   logger.info('successfully loaded kube specs')
-  await customResourceManager.manageCrd({ customResourceManifest })
+
+  kubeClient.addCustomResourceDefinition(customResourceManifest)
```
Flydiverny, line 37: This wires up kubeClients fluent api it was dones inside `customResourceManager` before.

moolen (approved):
I'm not super familiar with the helm v2 setup, but your changes LGTM!

Issue #657, opened 2021-03-15
Vulnerabilities inside kubernetes-external-secrets container


betterltn:
Hello,

Had scanned the latest kubernetes-external-secrets container image for common vulnerabilities

NAME      INSTALLED  FIXED-IN  VULNERABILITY   SEVERITY
cookie    0.4.0                CVE-2017-18589  High
dot-prop  4.2.1                CVE-2020-8116   High
editor    1.0.0                CVE-2015-0903   High
fresh     0.5.2                CVE-2013-1779   Low
ftp       0.3.10               CVE-1999-0082   High
ftp       0.3.10               CVE-1999-0201   Medium
lodash    4.17.19              CVE-2020-8203   High
lodash    4.17.19              CVE-2020-28500  Medium
lodash    4.17.19              CVE-2021-23337  High
rc        1.2.8                CVE-2014-1936   High
tar       4.4.13               CVE-2007-4476   High
y18n      4.0.0                CVE-2020-7774   High

Had found these. Does anyone know if these packages can be removed safely if the use-case  is limited to "aws secret manager"

moolen:
What kind of scanner did you use and what exact version did you test?

betterltn:
I used the anchore grype tool for scanning. 

https://github.com/anchore/grype 

Tried to run against latest release and it gave the above results.

moolen:
I did a quick analysis:

* false positive: cookie 0.4.0 CVE-2017-18589 (this is about a RUST package, not node)
* there's no fix available, not yet updated upstream: dot-prop 4.2.1  CVE-2020-8116
* false positive: editor 1.0.0 CVE-2015-0903
* false positive: ftp (CVE is about ftpd, not ftp npm package)
* waiting for fix from upstream: lodash 4.17.19

express is having the similar issues with these false-positives using anchore: https://github.com/expressjs/express/issues/4171#issuecomment-739585120

Flydiverny:
> there's no fix available, not yet updated upstream: dot-prop 4.2.1 CVE-2020-8116

As far as I can see we resolve a dot-prop version of 5.2.0 which is not one of the affected versions 🤔 

> lodash 4.17.19 CVE-2020-8203 High
> waiting for fix from upstream: lodash 4.17.19

Not sure I'm reading this correctly, but 4.17.19 is the patched version for the prototype pollution. This however was for zipObjectDeep which is not something we've used. 
[...]

PR #667, opened 2021-03-18
fix: update transitive deps

Flydiverny:
Signed-off-by: Markus Maga <markus@maga.se>

This is a full update of package-lock.json updating all deps within allowed dependency ranges.
Somewhat related to #657


moolen (approved):
LGTM

PR #681, opened 2021-03-29
Update crd

Flydiverny:
Fixes #680, #645, #621, #377

PR #682, opened 2021-03-29
fix!: require .spec field in CRD validation

Flydiverny:
Cherry picks from #681 

- fix invalid CRD test
- add test for template.type


Flydiverny:
Wanted to see if test would pass on master. but as expected no it doesn't because the validation schema has been somewhat loose. Setting spec as required e3cd9df would probably pass 🤔

Issue #697, opened 2021-04-07
Helm 2 EOL: chart apiVersion should be bumped?


sfynx:
When installing the chart with Argo CD, it uses Helm 2 when the chart `apiVersion` is `v1`, and Helm 3 when it is `v2`. As it is still `v1` at this time, this causes it to ignore the `crds/` directory.

https://github.com/external-secrets/kubernetes-external-secrets/blob/661cce1a344e019294f93c243c030a5f76e9ae0f/charts/kubernetes-external-secrets/Chart.yaml#L1

Everything looks great when changing it to `v2`, which should be fine if we don't expect people to use Helm 2 anymore.

Flydiverny:
#698 was merged

PR #698, opened 2021-04-07
fix: bump Helm chart API version

sfynx:
This addresses #697. Lemme know if I overlooked something.


Flydiverny (approved):
Thanks, missed this when cutting our helm v2 :) will merge when e2e passes

PR #701, opened 2021-04-09
feat(Multitenancy): scope KES access using ExternalSecret config

aabouzaid:
As discussed in PR #549 but using the `spec.controllerId` instead of the annotation as @Flydiverny suggested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,7 +435,7 @@ kind: ExternalSecret
 metadata:
   name: foo
 spec:
-  controllerID: 'dev-team-instance'
+  controllerId: 'dev-team-instance'
 [...]
 ```
 
--- a/README.md
+++ b/README.md
@@ -361,19 +361,6 @@ $ kubectl get secret/tmpl-ext-sec -ogo-template='{{ .metadata.labels }}'
 map[label1:11 label2:hello-world]
 ```
 
-## Disabling custom resource manager
-
-In case there is more than one kubernetes-external-secrets deployment, it's recommended to disable CRD manager
-to avoid having multiple deployments fighting over the CRD.
-
-That's could be done in the controller side by setting the env var:
-```yaml
-env:
-  DISABLE_CUSTOM_RESOURCE_MANAGER: true
-```
-
-Or in Helm, by setting `customResourceManagerDisabled=true`.
-
 ## Scoping access
 
 ### Using Namespace annotation
--- a/README.md
+++ b/README.md
@@ -361,6 +361,19 @@ $ kubectl get secret/tmpl-ext-sec -ogo-template='{{ .metadata.labels }}'
 map[label1:11 label2:hello-world]
 ```
 
+## Disabling custom resource manager
+
+In case there is more than one kubernetes-external-secrets deployment, it's recommended to disable CRD manager
+to avoid having multiple deployments fighting over the CRD.
+
+That's could be done in the controller side by setting the env var:
+```yaml
+env:
+  DISABLE_CUSTOM_RESOURCE_MANAGER: true
+```
+
+Or in Helm, by setting `customResourceManagerDisabled=true`.
+
 ## Scoping access
 
 ### Using Namespace annotation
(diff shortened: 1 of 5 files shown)


aabouzaid:
One side note here, on the controller side, I'm not sure if `Instance ID` is a good name or it should be `Controller ID` ... I just found it a bit confusing if it has the same name as in the `ExternalSecret` config. 

I have no strong opinion about it and could change it.

lib/daemon.js
```diff
@@ -62,6 +64,16 @@ class Daemon {
    */
   async start () {
     for await (const event of this._externalSecretEvents) {
+      // Check if the externalSecret should be managed by this instance.
+      if (event.object.spec) {
+        const externalSecretMetadata = event.object.metadata
+        const externalSecretController = event.object.spec.controllerID
+        if (externalSecretController && this._instanceId !== externalSecretController) {
```
Flydiverny, line 71: So in this case if an ExternalSecret does not specify controllerId an instance which has an instanceId would still look at it. 
Maybe we should modify it like this:

```suggestion
        if ((this._instanceId || externalSecretController) && this._instanceId !== externalSecretController) {
```

So if an KES instance has an instanceId the instance would only deal with those with matching controllerId, so if you want a KES instance taking care of all those without controller id you would leave instance id empty.

I suppose that boils down to just comparing in the end 😄 
```suggestion
        if (this._instanceId !== externalSecretController) {
```

lib/daemon.js
```diff
@@ -62,6 +64,16 @@ class Daemon {
...
+          this._logger.debug('the secret %s/%s is not managed by this instance but by %s',
+            externalSecretMetadata.namespace, externalSecretMetadata.name, externalSecretController)
+        }
```
Flydiverny, line 74: Looks like we are missing some skipping logic, now it would just log but then handle it anyways?

```suggestion
           continue
        }
```

or am I missing something here?

lib/daemon.test.js
```diff
@@ -106,4 +107,61 @@ describe('Daemon', () => {
...
+
+    expect(daemon._addPoller.called).to.equal(true)
+    expect(daemon._removePoller.calledWith('test-id')).to.equal(true)
+  })
+
+  it('do not manage externalsecrets with unmatched controller id and instance id', async () => {
+    const fakeExternalSecretEvents = (async function * () {
+      yield {
+        object: {
+          type: 'ADDED',
```
Flydiverny, line 144: Looks like these lines are mixed up causing the test to pass.
```suggestion
        type: 'ADDED',
        object: { 
```

lib/daemon.js
```diff
@@ -62,6 +64,16 @@ class Daemon {
    */
   async start () {
     for await (const event of this._externalSecretEvents) {
+      // Check if the externalSecret should be managed by this instance.
+      if (event.object.spec) {
+        const externalSecretMetadata = event.object.metadata
+        const externalSecretController = event.object.spec.controllerID
```
Flydiverny, line 70: Maybe we can case this `controllerID` like `controllerId` ?

Flydiverny (commented):
Thanks for taking the time! 🎉 
Regarding instance id vs controller id no really opinion here either, we can do like this it makes sense to me 😄

aabouzaid:
> Thanks for taking the time! tada
> Regarding instance id vs controller id no really opinion here either, we can do like this it makes sense to me smile

Thanks a lot for your feedback :bow: 
All of your comments have been addressed :heavy_check_mark:

README.md
```diff
@@ -361,6 +361,19 @@ $ kubectl get secret/tmpl-ext-sec -ogo-template='{{ .metadata.labels }}'
...
+to avoid having multiple deployments fighting over the CRD.
+
+That's could be done in the controller side by setting the env var:
+```yaml
+env:
+  DISABLE_CUSTOM_RESOURCE_MANAGER: true
+```
+
+Or in Helm, by setting `customResourceManagerDisabled=true`.
+
```
Flydiverny, line 376: This was removed in 7.0.0
```suggestion
```

Flydiverny (approved):

PR #708, opened 2021-04-14
Fix startup crash and ignore vscode config

Aenima4six2:
The check to see if a externalSecret should be managed by the running instance is not properly checking if `event.object` is present before evaluating the `spec` field.


Flydiverny (approved):
Thanks for fixing this :D didn't get any time yet

PR #710, opened 2021-04-16
Support path for vault-backend

Aenima4six2:
This pull request adds support for the `path` parameter of the existing `ExternalSecret` CRD.

Providing the path would previously throw an exception since there was no implementation of `_getByPath`.  

The following is an example using the `path` parameter.
```
apiVersion: "kubernetes-client.io/v1"
kind: ExternalSecret
metadata:
  name: hello-vault-service
spec:
  backendType: vault
  data:
    - path: secrets/data/foo # This will extract all keys from the secret foo
```

The implementation is quite simple. Since the existing `_get` method returns the entire Vault secret, serialized as JSON, we can reuse all the logic here, and simply return  back the object representation of the secret. The `kv-backend` will then pluck all keys from this object.

### DIFF
--- a/lib/backends/vault-backend.js
+++ b/lib/backends/vault-backend.js
@@ -99,6 +99,25 @@ class VaultBackend extends KVBackend {
 
     throw new Error('Unknown "kvVersion" specified')
   }
+
+  /**
+   * Get secret property value from Vault.
+   * @param {string} path - Path to the Secret in the backend.
+   * @param {object} keyOptions - Options for this specific key, eg version etc.
+   * @param {object} specOptions - Options for this external secret, eg role
+   * @param {string} specOptions.vaultMountPoint - mount point
+   * @param {string} specOptions.vaultRole - role
+   * @param {number} specOptions.kvVersion - K/V Version 1 or 2
+   * @returns {Promise} Promise object representing secret property values.
+   */
+  async _getByPath ({ path, specOptions }) {
+    // The vault "key" in _get is actually the entire secret, so we can reuse this method.
+    const secretString = await this._get({ key: path, specOptions })
+    // this isn't great, but that's the way it goes w/ inheritance.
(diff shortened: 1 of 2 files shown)


Flydiverny:
I'm not sure I understand the use case here. If you use `path` there shouldn't be a `name` so where does the value end up?
`path` is for fetch all secret keys in a given path. So if you have `secrets/data/foo/bar` and `secrets/data/foo/baz` and enter `secrets/data/foo` you would get both bar and baz resolved.
If you just want the path merged into the spec you should use `dataFrom`  and if you want the whole object at `secrets/data/foo` into `name: foo` then you should leave the `property` selector out.

Aenima4six2:
> I'm not sure I understand the use case here. If you use `path` there shouldn't be a `name` so where does the value end up?
> 
> `path` is for fetch all secret keys in a given path. So if you have `secrets/data/foo/bar` and `secrets/data/foo/baz` and enter `secrets/data/foo` you would get both bar and baz resolved.
> 
> If you just want the path merged into the spec you should use `dataFrom`  and if you want the whole object at `secrets/data/foo` into `name: foo` then you should leave the `property` selector out.

Apologies is this functionality is possible via another configuration. 'name' in my example should have been omitted, as it's not used. What we are trying achieve is render a direct mapping of all key/value pairs from a Vault secret to a K8s secret. So if a vault path (secret) at '/foo/bar' contains keys 'aaa' and 'bbb', with this patch, and specifying path '/foo/bar', you would have a K8s secret with the same keys, 'aaa' and 'bbb' with the respective values.

Flydiverny:
Sorry for not getting back here. Did you find something that solved your use case? :)

Aenima4six2:
> Sorry for not getting back here. Did you find something that solved your use case? :)

No problem at all. `dataFrom` was exactly what I was looking for. Somehow I missed understood what its use case was. Thanks for pointing me in the right direction.

PR #758, opened 2021-05-24
CRD property types

### LINKED ISSUE #753: CRD empty properties breaking intellij parsing (cause of empty properties)
Hi,

When including the CRD for parsing in the intellij, it breaks parsing as the CRD is having empty properties which are not supported there in openApiSchema.

The parsing will be fixed to not break in the intellij following the issue:
https://youtrack.jetbrains.com/issue/IDEA-269698

Probably the empty properties could just be omitted? (I don't know openApi well enough)
```yaml
                    oneOf:
                      - properties:
                          key:
                          name:
                          property:
                          isBinary:
                          secretType:
                        required:
                          - key
                          - name
                      - properties:
                          path:
                          recursive:
                        required:
                          - path
```

davidholsgrove:
Closes #753

The empty properties here also produces invalid json schema when used with https://github.com/yannh/kubeconform

### DIFF
--- a/charts/kubernetes-external-secrets/crds/kubernetes-client.io_externalsecrets_crd.yaml
+++ b/charts/kubernetes-external-secrets/crds/kubernetes-client.io_externalsecrets_crd.yaml
@@ -122,26 +122,10 @@ spec:
                           Used by: secretsManager
                         type: string
                     oneOf:
-                      - properties:
-                          key:
-                            type: string
-                          name:
-                            type: string
-                          property:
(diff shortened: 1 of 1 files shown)


caretak3r:
This is also currently an issue with us using the latest (8.0 and 8.0.1). Bump ^

davidholsgrove:
Could this PR (or any alternative to address #753 ) be reviewed / merged before the next cut of `external-secrets`?

Issue #763, opened 2021-05-26
Application stops functioning after a few days


jlact:
Application seems to just stop functioning, usually after a few days. No action is taken to stop or terminate deployment. As far as kubernetes is concerned the pod is still running, and it does indeed respond to calls on the metrics endpoint, just doesn't manage the secrets anymore. Issuing a rolling restart on the deployment resolves it, but it's a manual fix.

version: 7.2.0
k8s: 1.19
backend: Azure Keyvault

Last entries in log files look like:
```
{"level":"info","message_time":"2021-05-21T22:12:49.695Z","pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5bdc8f9947-8x7pg","message":"Stopping watch stream for namespace * due to event: END"}
{"level":"info","message_time":"2021-05-21T22:12:49.699Z","pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5bdc8f9947-8x7pg","message":"stopping poller for ns1/notifications-secret"}
{"level":"info","message_time":"2021-05-21T22:12:49.699Z","pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5bdc8f9947-8x7pg","message":"stopping poller for ns1/private-key"}
{"level":"info","message_time":"2021-05-21T22:12:49.699Z","pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5bdc8f9947-8x7pg","message":"stopping poller for ns2/dns-config"}
{"level":"info","message_time":"2021-05-21T22:12:49.699Z","pid":17,"hostname":"external-secrets-kubernetes-external-secrets-5bdc8f9947-8x7pg","message":"stopping poller for ns3/external-dns-config"}
```

Metrics from kubernetes show CPU (and network) usage as 0 once it occurs, memory usage is also static. Green being the 'failed' pod and orange a new one from restarting the deployment:
![image](https://user-images.githubusercontent.com/10630610/119648236-20484880-be19-11eb-8b97-cac3219e3995.png)

anderius:
Same problem here:

helm chart version: 8.0.1
k8s (AKS): 1.20.5
backend: Azure Keyvault

Flydiverny:
Should be resolved in 8.0.2 by #765

PR #765, opened 2021-06-02
Ensure that the restart timer is always started

battlecow:
Related: #763 

If there are no data events, the reset timer is never started.  If the connection to the kube api server is lost during this time and new events come in they are never seen and the stream is never restarted.

Repro: (I used minikube with ssh access)
Install helm chart
`minikube ssh`
Drop current connection from pod to api server: e.g. `conntrack -D conntrack -s 172.17.0.6 -d 10.96.0.1`
Create a new example secret (failures to connect don't matter here)
Note that the secret is not picked up nor is the server restarted. `kubectl get externalsecrets.kubernetes-client.io`

Credit and thanks to https://github.com/external-secrets/kubernetes-external-secrets/issues/362#issuecomment-719989503 for the iptables idea.

### DIFF
--- a/lib/external-secret.js
+++ b/lib/external-secret.js
@@ -76,6 +76,8 @@ async function startWatcher ({
         clearTimeout(timeout)
       })
 
+      restartTimeout()
+
       const deathEvent = await deathQueue.take()
 
       logger.info('Stopping watch stream for namespace %s due to event: %s', loggedNamespaceName, deathEvent)


Flydiverny (approved):
Oh god, that makes a lot of sense 😄 
Thanks for finding this!

lib/external-secret.js
```diff
@@ -76,6 +76,8 @@ async function startWatcher ({
         clearTimeout(timeout)
       })
 
+      restartTimeout();
```
Flydiverny, line 79: Picky lint rules here
```suggestion
      restartTimeout()
```

battlecow:
This took me an embarrassingly long time to figure out and make sure I reproduced it correctly. =]

battlecow:
Did a rebase just to remove the silly semicolon removal

Issue #737, opened 2021-05-07
Upsert secret only when it has changed


guxlightyear:
I appreciate this might not be a feature useful for many people, but currently ExternalSecrets is calling `await kubeNamespace.secrets.post({ body: secretManifest })` (in `lib/poller.js#_upsertKubernetesSecret`), regardless of whether the kubernetes `secret` has actually changed or not.

This is a problem if you have many secrets in a cluster with many validation webhooks, since these webhooks are going to be invoked for each post operation, even if the resource already existed and had not changed.

We have encountered this in one of our clusters where we have 725 external secrets (with the default 10s polling interval), and are using [Gatekeeper](https://github.com/open-policy-agent/gatekeeper) with [CIS security policies](https://github.com/raspbernetes/k8s-security-policies/tree/master/policies).

After enabling the policies we noticed external-secrets was failing with

```
level":50,"time":1620370654942,"pid":18,"hostname":"platform-kubernetes-external-secrets-578c5b47f4-p65rk","code":429,"statusCode":429,"stack":"Error: Too many requests, please try again later.\n\n    at /app/node_modules/kubernetes-client/backends/request/client.js:231:25\n    at Request._callback (/app/node_modules/kubernetes-client/backends/request/client.js:168:14)\n    at Request.self.callback (/app/node_modules/kubernetes-client/node_modules/request/request.js:185:22)\n    at Request.emit (events.js:315:20)\n    at Request.EventEmitter.emit (domain.js:482:12)\n    at Request.<anonymous> (/app/node_modules/kubernetes-
```

The root cause was that our API server was getting flooded as a result of having roughly 70 requests to update secrets per second, each resulting in around 60 requests (one per constraint) validation requests to the API server.

We can certainly workaround our issue by:
* Being a lot more conservative with the validations we perform to objects in our cluster
* Reducing the polling period for our secrets ([polling per secret](https://github.com/external-secrets/kubernetes-external-secrets/issues/723) would be great for that)

But ideally this is fixed at the source, so that only secrets that have actually changed trigger a `post` request

PR #782, opened 2021-06-25
feat: Upsert secrets only when needed

stephenthedev:
This implements a rudimentary data check on secrets polled to only call put/post actions when the data of the secret has changed.

This is aimed at resolving #737

### DIFF
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -163,14 +163,57 @@ class Poller {
     }
 
     const secretManifest = await this._createSecretManifest()
-    this._logger.info(`upserting secret ${this._namespace}/${this._name}`)
+    const kubeSecret = kubeNamespace.secrets(this._name)
+    let existingSecret
 
     try {
-      return await kubeNamespace.secrets.post({ body: secretManifest })
+      this._logger.info(`getting secret ${this._namespace}/${this._name}`)
+      const secretResponse = await kubeSecret.get()
+      existingSecret = secretResponse.body
     } catch (err) {
-      if (err.statusCode !== 409) throw err
-      return kubeNamespace.secrets(this._name).put({ body: secretManifest })
+      if (err.statusCode !== 404) throw err
+      // do nothing if the secret is not found
     }
+
+    if (existingSecret && this._equalSecretData(existingSecret, secretManifest)) {
+      this._logger.info(`skipping secret ${this._namespace}/${this._name} upsert, objects are the same`)
+      return Promise.resolve(true)
+    } else if (existingSecret === undefined) {
+      this._logger.info(`creating secret ${this._namespace}/${this._name}`)
+      return await kubeNamespace.secrets.post({ body: secretManifest })
+    } else {
+      this._logger.info(`updating secret ${this._namespace}/${this._name}`)
+      return kubeSecret.put({ body: secretManifest })
+    }
+  }
+
+  /**
+   * Checks if a secret and the desired secret manifest are equal
+   *
+   * @param   {Object}  kubeSecret      An actual kubernetes secret from the kube-client
+   * @param   {Object}  secretManifest  A hash representing the desired secret
+   *
+   * @return  {Boolean}                  Boolean if they are the same or not
+   */
+  _equalSecretData (kubeSecret, secretManifest) {
+    let result = true
+    const liveSecret = clonedeep(kubeSecret)
+    const desiredSecret = clonedeep(secretManifest)
+
+    // Only use annotations and labels for metadata checking
+    const secrets = [liveSecret, desiredSecret]
+    secrets.forEach((s) => {
+      s.metadata = {
+        labels: s.metadata.labels,
+        annotations: s.metadata.annotations
+      }
+    })
+
+    result = result ? JSON.stringify(liveSecret.metadata) === JSON.stringify(desiredSecret.metadata) : false
+
+    // check secret data
+    result = result ? JSON.stringify(liveSecret.data) === JSON.stringify(desiredSecret.data) : false
+    return result
   }
 
   async _updateStatus (status) {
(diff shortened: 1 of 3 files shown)


stephenthedev:
@Flydiverny Can you help kick off CI? Thanks!

lib/poller.js
```diff
@@ -163,16 +163,42 @@ class Poller {
...
+  /**
+   * Checks if a secret and the desired secret manifest are equal
+   *
+   * @param   {Object}  kubeSecret      An actual kubernetes secret from the kube-client
+   * @param   {Object}  secretManifest  A hash representing the desired secret
+   *
+   * @return  {Boolean}                  Boolean if they are the same or not
+   */
+  _equalSecretData (kubeSecret, secretManifest) {
+    return JSON.stringify(kubeSecret.data) === JSON.stringify(secretManifest.data)
```
moolen, line 199: I think here is an edge-case when a user renders secret-data into the `metadata` object (via the `spec.template` field). The secret wouldn't change then ONLY IF the pulled secret is not part of the `secret.data`.

stephenthedev: Thanks @moolen good catch, I updated the function to consider custom defined metadata. Please check it out an LMK!

stephenthedev: I also added a test for this scenario as well 👍🏽

moolen:
LGTM so far, thank you! I want to check out your branch and do some manual testing.

stephenthedev:
@moolen awesome thanks! Looking forward to this landing for our use case :-)

lib/poller.js
```diff
@@ -163,16 +163,60 @@ class Poller {
...
+    const liveMetadata = clonedeep(kubeSecret.metadata)
+    const filterList = [
+      'creationTimestamp',
+      'namespace',
+      'ownerReferences',
+      'resourceVersion',
+      'selfLink',
+      'uid'
+    ]
+    filterList.forEach(k => delete liveMetadata[k])
```
moolen, line 211: This `filterList` needs to be applied to the `kubeSecret` (the current data) aswell when comparing the stringified json.

moolen (changes requested):
I tested it locally but found that it wasn't working because some `metadata` fields were present:

```json
// new metadata (from `secretManifest`)
{
    "labels": {
        "secretLabel": "Hellofoo123"
    },
    "name": "e2e-secretmanager-template-sql65h62hxarnqjmpgkj4",
    "ownerReferences": [
        {
            "apiVersion": "kubernetes-client.io/v1",
            "controller": true,
            "kind": "ExternalSecret",
            "name": "e2e-secretmanager-template-sql65h62hxarnqjmpgkj4",
            "uid": "b13c2a96-5258-4ec5-be37-7217aec71220"
        }
    ]
}
// current metadata (from `kubeSecret`)
{
    "name": "e2e-secretmanager-template-sql65h62hxarnqjmpgkj4",
    "labels": {
        "secretLabel": "Hellofoo123"
    },
    "managedFields": [
        {
            // ...
        }
    ]
}
```

I suggested two things in the comments. it would be great to have e2e tests for that (i did them manually today) but i don't have an approach for your right now that from the top of my head. Maybe you can figure something out if you have the time for it.

stephenthedev:
@moolen made the change. As for the e2e test, we don't have any existing instrumentation to trap calls made to the API so that'd be a bit of a lift to assert some of that. Given its now an allow list, I think the vector for that becomes pretty small. What do you think?

stephenthedev:
Thinking about this more, the allowList also makes the unit test a lot more accurate in asserting the behavior, so maybe the e2e is overkill?

lib/poller.js
```diff
@@ -163,14 +163,57 @@ class Poller {
...
+    // Only use annotations and labels for metadata checking
+    const secrets = [liveSecret, desiredSecret]
+    secrets.forEach((s) => {
+      s.metadata = {
+        labels: s.metadata.labels,
+        annotations: s.metadata.annotations
+      }
+    })
+
+    result = result ? JSON.stringify(secretManifest.metadata) === JSON.stringify(kubeSecret.metadata) : false
```
moolen, line 212: you need to compare `liveSecret` and `desiredSecret`!

lib/poller.js
```diff
@@ -163,14 +163,57 @@ class Poller {
...
+
+    // check secret data
+    result = result ? JSON.stringify(kubeSecret.data) === JSON.stringify(secretManifest.data) : false
```
moolen, line 215: here, too

lib/poller.test.js
```diff
@@ -696,18 +701,84 @@ describe('Poller', () => {
       })).to.equal(true)
     })
 
+    it('doesn\'t update a secret if it hasn\'t changed', async () => {
```
moolen, line 704: This test can't really check whether it has been updated, patched or nothing happened. Let's create a test to explicitly test `_equalSecretData()` instead to ensure it returns the correct value for the given arguments.

moolen:
> Given its now an allow list, I think the vector for that becomes pretty small. What do you think?

Sure, that works for me!

lib/poller.js
stephenthedev: oh my, so sorry for missing it 😅 updated!

lib/poller.js
stephenthedev: oh my, so sorry for missing it 😅 updated!

lib/poller.test.js
stephenthedev: Added the assertions to not have put or post called

moolen (approved):
LGTM! Thanks for your contribution :tada:

stephenthedev:
@moolen Awesome thank you for the review! What is the release cadence for this?

Flydiverny:
@stephenthedev I made a release now :)
Thanks for the contribution!

stephenthedev:
Amazing thank you both!

PR #790, opened 2021-07-09
Add 'dataFromLiteral' to ExternalSecret API

jwtracy:
I have a use case where I need a Kubernetes secret which contains a file
meant to be mounted to a container. In this file there is sensitive data
so it must be templated using the `spec.template.stringData`
functionality. Within this file there is also recurring, non-sensitive,
data that would benefit from being templated.

More specifically, I had a dozen references to a GCP project ID which
would need to be hard coded in that many locations.

I tried to supply this within the `spec.template.stringData`, but
expectedly, that data is not part of the map which renders the template;
only `spec.data` and `spec.dataFrom` are merged for that.

It would be possible to use Helm to fill Golang templates within the
lodash template, however, if feels like a bad practice to have 2
templating patterns operating on the same text.

This led me to introduce the `dataFromLiteral` key-value pairing for the
abstract kvbackend so all backends can template non-sensitive data along
side their sensitive, cloud secrets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,9 @@ spec:
       name: s1
     - key: kv/data/extsec/secret2
       name: s2
+  dataFromLiteral:
+    - name: s3
+      value: non-sensitive
   kvVersion: 2
   template:
     data:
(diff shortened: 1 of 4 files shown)


javier-b-perez (approved):

jwtracy:
https://github.com/external-secrets/kubernetes-external-secrets/issues/835

PR #794, opened 2021-07-12
fix: remove AWS_DEFAULT_REGION

Flydiverny:
Not sure why AWS_DEFAULT_REGION was added in the first place, but it seems misleading to have. Its not provided in the docs everywhere and the AWS SDK for JS uses `AWS_REGION` so the `AWS_DEFAULT_REGION` should not be used, if one uses the aws cli it does seem to prefer `AWS_DEFAULT_REGION` though in some cases 😄

### DIFF
--- a/charts/kubernetes-external-secrets/values.yaml
+++ b/charts/kubernetes-external-secrets/values.yaml
@@ -5,7 +5,6 @@
 # Environment variables to set on deployment pod
 env:
   AWS_REGION: us-west-2
-  AWS_DEFAULT_REGION: us-west-2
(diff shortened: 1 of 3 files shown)


moolen (approved):
That makes sense, this is a big inconsistency in the aws sdks.

arischow:
@Flydiverny 

I think after this change, there should be a new Helm chart version release? Well I am new to this repo but I have to say after installing this chart, this environment variable did confuse me, especially I specified the region (`AWS_REGION`) to another one.

PR #800, opened 2021-07-23
Fixes naming convention permission check for data items with path attribute.

vladlosev:
The permission checking code is currently ignoring the `path` attributes when evaluating. This results in permission checks for secrets defined like this:
```
kind: ExternalSecret
metadata:
  name: test
spec:
  backendType: systemManager
  data:
  - path: dev/team-a/secret
```
with the error message `ERROR, not allowed to fetch secret: test/test: key name undefined does not match naming convention ^dev/team-a/.*`.
This PR adds handling of the `path` attribute in the permissions check.

### DIFF
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -289,7 +289,8 @@ class Poller {
               allowed, reason
             }
           }
-        } else if (!reNaming.test(secretProperty.key)) {
+        }
+        if (!reNaming.test(secretProperty.key)) {
           allowed = false
           reason = `key name ${secretProperty.key} does not match naming convention ${namingConvention}`
           return {
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -284,7 +284,7 @@ class Poller {
         if (secretProperty.path) {
           if (!reNaming.test(secretProperty.path)) {
             allowed = false
-            reason = `path name ${secretProperty.path} does not match naming convention ${namingConvention}`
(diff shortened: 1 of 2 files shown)


vladlosev:
@Flydiverny can you please take a look at this? The problem this fixes is also a security vulnerability where the access check is performed on the `key` attribute whereas the values are fetched using the `path` attribute.

lib/poller.js
```diff
@@ -281,7 +281,15 @@ class Poller {
...
-        if (!reNaming.test(secretProperty.key)) {
+        if (secretProperty.path) {
+          if (!reNaming.test(secretProperty.path)) {
+            allowed = false
+            reason = `path ${secretProperty.path} does not match naming convention ${namingConvention}`
+            return {
+              allowed, reason
+            }
+          }
+        } else if (!reNaming.test(secretProperty.key)) {
```
Flydiverny, line 292: While this shouldn't happen due to both CRD validation, and then later the backend implementation shouldn't use both path and key. It feels better to have them both validated and not just path if they for some reason both exist (like in the unit test)

```suggestion
        }
        if (!reNaming.test(secretProperty.key)) {
```

Flydiverny (commented):
Sorry for the wait, just one minor thought

lib/poller.js
vladlosev: Done.

Flydiverny (approved):

PR #830, opened 2021-09-09
Fixes naming convention permission check for data items with path attribute only.

vladlosev:
This change fixes the false positive in the permission checking code when the `path` attribute is used. Without the fix, the added test fails, as the as the code always tries to check the `key` attribute, even when only `path` is provided. It ends up coercing the `undefined` value to a sting and testing that.

### DIFF
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -289,8 +289,7 @@ class Poller {
               allowed, reason
             }
           }
-        }
-        if (!reNaming.test(secretProperty.key)) {
+        } else if (!reNaming.test(secretProperty.key)) {
           allowed = false
           reason = `key name ${secretProperty.key} does not match naming convention ${namingConvention}`
           return {
--- a/lib/poller.js
+++ b/lib/poller.js
@@ -281,15 +281,15 @@ class Poller {
     // Testing data property
     if (namingConvention && externalData) {
       externalData.forEach((secretProperty, index) => {
-        if (secretProperty.path) {
-          if (!reNaming.test(secretProperty.path)) {
-            allowed = false
-            reason = `path ${secretProperty.path} does not match naming convention ${namingConvention}`
-            return {
-              allowed, reason
-            }
+        if ('path' in secretProperty && !reNaming.test(secretProperty.path)) {
+          allowed = false
+          reason = `path ${secretProperty.path} does not match naming convention ${namingConvention}`
+          return {
+            allowed, reason
           }
-        } else if (!reNaming.test(secretProperty.key)) {
+        }
+
+        if ('key' in secretProperty && !reNaming.test(secretProperty.key)) {
           allowed = false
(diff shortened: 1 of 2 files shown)


lib/poller.js
```diff
@@ -289,8 +289,7 @@ class Poller {
               allowed, reason
             }
           }
-        }
-        if (!reNaming.test(secretProperty.key)) {
+        } else if (!reNaming.test(secretProperty.key)) {
```
Flydiverny, line 292: I'm thinking its better we check for the respective keys and validate them if present.

```js
      externalData.forEach((secretProperty, index) => {
        if ('path' in secretProperty && !reNaming.test(secretProperty.path)) {
          allowed = false
          reason = `path ${secretProperty.path} does not match naming convention ${namingConvention}`
          return {
            allowed, reason
          }
        }

        if ('key' in secretProperty && !reNaming.test(secretProperty.key)) {
          allowed = false
          reason = `key name ${secretProperty.key} does not match naming convention ${namingConvention}`
          return {
            allowed, reason
          }
        }
      })
```

They never should be present at the same time, but if validation or other layer of the code allows for it, then this would potentially allow bypassing the validation. By supplying a bogus path that would validate and using a key which is ignored by validation.

lib/poller.test.js
```diff
@@ -982,6 +982,16 @@ describe('Poller', () => {
...
+        {
+          // test regex on path
+          ns: { metadata: { annotations: { [namingPermittedAnnotation]: 'dev/team-a/.*' } } },
+          descriptor: {
+            data: [
+              { path: 'dev/team-a/secret' }
+            ]
+          },
+          permitted: true
+        },
```
Flydiverny, line 994:
```suggestion
        },
        {
          ns: { metadata: { annotations: { [namingPermittedAnnotation]: 'dev/team-a/.*' } } },
          descriptor: {
            data: [
              { key: 'dev/team-a/secret', name: 'somethingelse', path: '' }
            ]
          },
          permitted: false
        },
        {
          ns: { metadata: { annotations: { [namingPermittedAnnotation]: 'dev/team-a/.*' } } },
          descriptor: {
            data: [
              { key: 'this-should-fail', name: 'somethingelse', path: 'dev/team-a/such-path' }
            ]
          },
          permitted: false
        },
        {
          ns: { metadata: { annotations: { [namingPermittedAnnotation]: 'dev/team-a/.*' } } },
          descriptor: {
            data: [
              { key: 'dev/team-a/such-key', name: 'somethingelse', path: 'this-should-fail' }
            ]
          },
          permitted: false
        },
```

lib/poller.js
vladlosev: Yeah, checking `path` and `key` independently makes sense. I updated the code to do that.

Flydiverny (approved):

Issue #864, opened 2021-11-02
Announcement: Take a look at ESO 


Flydiverny:
Hi! 👋 

Are you a new user looking to start using KES? 
Please take a look at the ESO instead!

**KES** = Kubernetes Extenral Secrets (this repo)
**ESO** = External Secrets Operator (repo: [external-secrets/external-secrets](https://github.com/external-secrets/external-secrets))

KES is mainly under maintenance and the go rewrite ESO has come a long way and most likely covers your needs!
If it doesn't please raise an issue over there or even better contribute!

ESO offers better documentation, better configuration possibilities and uses modern dependencies 😄

ESO can be found here:
https://external-secrets.io/
https://github.com/external-secrets/external-secrets

KES does not have any dedicated or active contributions at this time and is mainly on life support (and has been for awhile). As can be seen from some issues, there are tech debts in this project that are causing issues and some of those dependencies are no longer maintained. To replace them would have required a sizeable effort and its even harder to motivate now with ESO having come a long way!

If you have any questions or so feel free to join the community over at [#external-secrets](https://kubernetes.slack.com/archives/C017BF84G2Y) in the Kubernetes slack!

derekwilliamsliquidx:
This is an acceptable solution to my issues. While it's annoying to have to replace the KES engine, this new project looks promising.

thanks!

Flydiverny:
There's a tool WIP to ease migration path:
https://github.com/external-secrets/kes-to-eso
But I think most of it is still in a PR https://github.com/external-secrets/kes-to-eso/pull/1 but we'd love help testing it!

vgazzola:
I think a deprecation notice on the main README is needed :)
https://github.com/external-secrets/kubernetes-external-secrets/pull/905

muffl0n:
Would love to have the deprecation notice on artifact hub: https://artifacthub.io/packages/helm/external-secrets/kubernetes-external-secrets

Flydiverny:
> Would love to have the deprecation notice on artifact hub: [artifacthub.io/packages/helm/external-secrets/kubernetes-external-secrets](https://artifacthub.io/packages/helm/external-secrets/kubernetes-external-secrets)

Should be fixed in 8.5.3 might take awhile before it shows up on artifacthub tho

muffl0n:
Ah, there it is! Thank you! 😍

sufyanadam:
[...]

PR #868, opened 2021-11-11
Log base w/ configurable pid and hostname keys

SocalNick:
`hostname` isn't always desireable and is somewhat inaccurate given
os.hostname is the pod name in kubernetes


SocalNick:
👋 folks! My company is using k8s-external-secrets...thanks so much for developing / maintaining it!

The [pino default log base](https://getpino.io/#/docs/api?id=base-object) uses hostname as the key for `os.hostname()`. In a kubernetes context, this evaluates to the pod name. When our log aggregation platform processes these logs w/ hostname set to the pod name, it can't decorate the logs w/ additional tags for the actual underlying host. This change maintains the default behavior but allows an operator to choose alternate keys for the pid and hostname of the pino log base.

I didn't see any tests around configuration, but if this feels like it requires a test, happy to add one if you can point me in the right direction.

Also, this likely needs to be documented. Can you point me towards the docs that I should update?


Thanks!

Nick

SocalNick:
@Flydiverny @riccardomc any chance you could take a look at this PR? Thanks!

Flydiverny (approved):
Also see #864

Flydiverny:
Regarding docs it would be here:
https://github.com/external-secrets/kubernetes-external-secrets/blob/master/charts/kubernetes-external-secrets/README.md#configuration

SocalNick:
Added documentation, thanks again for your consideration!

Flydiverny:

Issue #826, opened 2021-08-23
status check went boom


djfinnoy:
Possibly same issue as #362.

Running v8.3.0, installed via Helm chart, on GKE (1.20.8-gke.900).
The cluster gets bootstrapped by an automated pipeline, and `external-secrets` is expected to retrieve a single secret  from Google Secret Manager.
I works about 50% of the time.

When it doesn't work, I'm seeing the following lines repeated in the pod logs:
```
{"level":50,"message_time":"2021-08-23T08:31:59.073Z","pid":18,"hostname":"external-secrets-kubernetes-external-secrets-b44fd8967-dgplh","payload":{},"msg":"status check went boom for oauth2-proxy/sm-oauth-github"}
{"level":30,"message_time":"2021-08-23T08:32:59.123Z","pid":18,"hostname":"external-secrets-kubernetes-external-secrets-b44fd8967-dgplh","msg":"No watch event for 60000 ms, restarting watcher for *"}
{"level":30,"message_time":"2021-08-23T08:32:59.123Z","pid":18,"hostname":"external-secrets-kubernetes-external-secrets-b44fd8967-dgplh","msg":"Stopping watch stream for namespace * due to event: END"}
{"level":30,"message_time":"2021-08-23T08:32:59.131Z","pid":18,"hostname":"external-secrets-kubernetes-external-secrets-b44fd8967-dgplh","msg":"starting poller for oauth2-proxy/sm-oauth-github"}
{"level":50,"message_time":"2021-08-23T08:32:59.131Z","pid":18,"hostname":"external-secrets-kubernetes-external-secrets-b44fd8967-dgplh","payload":{},"msg":"status check went boom for oauth2-proxy/sm-oauth-github"
```

Deleting the pod solves the problem, but I have not checked whether the problem re-occurs.
I never saw this problem before upgrading from 6.2.0 to 8.3.0.

Flydiverny:
> This happened again at 10:01am
> 
> ```
> {"level":30,"message_time":"2021-10-27T18:20:53.906Z","pid":17,"hostname":"kubernetes-external-secrets-69c544c96d-tkncr","msg":"No watch event for 60000 ms, restarting watcher for *"}
> ```

[...]

PR #877, opened 2021-12-07
feat: add runtime metrics

moolen:
Small enhancement for #826: export default `nodejs_*` metrics. They can be displayed with the [nodejs app dashboard](https://grafana.com/grafana/dashboards/11159).

### DIFF
--- a/bin/daemon.js
+++ b/bin/daemon.js
@@ -56,6 +56,7 @@ async function main () {
   })
 
   const registry = Prometheus.register
+  Prometheus.collectDefaultMetrics({ register: registry })
   const metrics = new Metrics({ registry })
 
   const pollerFactory = new PollerFactory({


bin/daemon.js
```diff
@@ -56,6 +56,7 @@ async function main () {
   })
 
   const registry = Prometheus.register
+  Prometheus.collectDefaultMetrics({ register: registry })
```
Flydiverny, line 59: Should we use the prefix option to set a prefix?
Other metrics start with `kubernetes_external_secrets_<some_metric_name>`

moolen: I would prefer using the default metric names. This way you can have a generic dashboard for nodejs apps like the one above.

Flydiverny: But wouldn't it conflict if you have other services exposing the same metrics? I'm a bit out of the metrics loop so maybe its a non issue 😄 I just know we got PRs before adding the prefix to follow Prometheus best practices

moolen: No, it wouldn't conflict, each instance gets its own set of labels.

Flydiverny (approved):

PR #881, opened 2021-12-09
Fetch all Azure KeyVault secrets

### LINKED ISSUE #880: Fetch all Azure KeyVault secrets
When many secrets are maintained in a single Azure KeyVault, it becomes rather cumbersome having to specify each secret twice, once in the KeyVault itself and again on the manifest for kubernetes-external-secrets.
It would be ideal if we could map a predetermined path (e.g. `secrets`) to the entire list of secrets. So they are added/removed dynamically.

monti-python:
closes #880


monti-python:
Apologies for linter issues, `npm test` passing now

PR #884, opened 2021-12-16
Add a more appropriate' history' heading

sbose78:
Now that it's been a while, I thought it would be better if we didn't have the 'Repository moved to external-secrets' text as the first thing which someone read :)


Flydiverny (approved):
Fair enough! 😄

PR #894, opened 2022-01-09
Expand SSM Parameter Store docs to clarify use of version constraints

oaklees:
Expands on the AWS SSM Parameter Store docs with examples of how to fetch a specific version of a given parameter.

Whilst this is not technically a feature offered by the package directly and instead supported by the underlying AWS SDK, it's a very useful feature worth highlighting which allows immutable, versioned controlled references to SSM instead of always fetching the latest value that can change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,7 +581,7 @@ spec:
       recursive: false
 ```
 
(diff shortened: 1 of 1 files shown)


Flydiverny (approved):

Flydiverny:
Thanks!

PR #927, opened 2022-05-23
Randomize

erikbotta:


Flydiverny:
While I think this is a good change, and something I wanted to add before but never remembered 😄 

Please see KES deprecation notice
- #864

erikbotta:
Yeah we are facing rate-limit at Hashicorp Vault so we need some random requests to their api to be good with 60 rps.

